PR #4, opened 2022-01-06
Deal with absolute paths in test output comparisons etc.

brson:
Also make the test harness `--bless` argument work.

This fixes the issue with `cargo test` failing that I mentioned in https://github.com/dada-lang/dada/pull/3. It's a hack. I understand if you want to fix it some other way.


nikomatsakis (approved):

nikomatsakis:
Makes sense. I've only ever run the test harness with `cargo run test --bless`, I realized. Seems good!

PR #13, opened 2022-01-06
Add GitHub actions workflows and bors.toml

### LINKED ISSUE #7: bors
We need [bors-ng](https://github.com/bors-ng/bors-ng)! And then I need to stop pushing directly to master, probably.

mbrobbel:
An attempt to close #7.

I also tried to fix Clippy warnings but now compilation fails with an ICE:

```
query stack during panic:
#0 [type_of] computing type of `lex::StringFormatBuffer::sections`
#1 [inferred_outlives_crate] computing the inferred outlives predicates for items in this crate
end of query stack
error: could not compile `dada-lex`
```

### DIFF
--- a/components/dada-execute/src/permission/my.rs
+++ b/components/dada-execute/src/permission/my.rs
@@ -43,12 +43,14 @@ impl My {
 
     pub(super) fn check_read(&self, interpreter: &Interpreter) -> eyre::Result<()> {
         self.given.check_still_valid(interpreter)?;
-        Ok(self.tenant.cancel_tenant_if_exclusive(interpreter))
+        self.tenant.cancel_tenant_if_exclusive(interpreter);
+        Ok(())
     }
 
     pub(super) fn check_write(&self, interpreter: &Interpreter) -> eyre::Result<()> {
         self.given.check_still_valid(interpreter)?;
-        Ok(self.tenant.cancel_tenant(interpreter))
+        self.tenant.cancel_tenant(interpreter);
+        Ok(())
     }
 
     pub(crate) fn check_await(&self, interpreter: &Interpreter) -> eyre::Result<()> {
--- a/components/dada-ir/src/span.rs
+++ b/components/dada-ir/src/span.rs
@@ -76,6 +76,7 @@ impl Span {
     }
 
     /// Returns a 0-length span at the start of this span
+    #[must_use]
     pub fn span_at_start(self) -> Span {
         Span {
             start: self.start,
@@ -94,6 +95,11 @@ impl Span {
         self.end - self.start
     }
 
+    pub fn is_empty(self) -> bool {
+        self.len() == 0
+    }
+
+    #[must_use]
     pub fn to(self, other: Span) -> Span {
         assert!(self == other || self.end <= other.start);
         Span {
@@ -141,14 +147,14 @@ impl From<u32> for Offset {
     }
 }
 
-impl Into<u32> for Offset {
-    fn into(self) -> u32 {
-        self.0
+impl From<Offset> for u32 {
+    fn from(offset: Offset) -> Self {
+        offset.0
     }
 }
 
-impl Into<usize> for Offset {
-    fn into(self) -> usize {
-        self.0 as usize
+impl From<Offset> for usize {
+    fn from(offset: Offset) -> Self {
+        offset.0 as usize
     }
 }
--- a/components/dada-lang/src/test_harness/lsp_client.rs
+++ b/components/dada-lang/src/test_harness/lsp_client.rs
@@ -65,10 +65,9 @@ impl ChildSession {
     fn send_any(&mut self, msg: impl Serialize) -> eyre::Result<()> {
         let msg_raw = serde_json::to_string(&msg)?;
 
-        let child_stdin = self.child.stdin.as_mut().ok_or(std::io::Error::new(
-            std::io::ErrorKind::BrokenPipe,
-            "can connect to child stdin",
-        ))?;
+        let child_stdin = self.child.stdin.as_mut().ok_or_else(|| {
+            std::io::Error::new(std::io::ErrorKind::BrokenPipe, "can connect to child stdin")
+        })?;
 
         child_stdin
             .write_all(format!("Content-Length: {}\r\n\r\n", msg_raw.len()).as_bytes())
@@ -90,25 +89,27 @@ impl ChildSession {
     }
 
     fn receive<T: for<'de> Deserialize<'de>>(&mut self) -> eyre::Result<T> {
-        let child_stdout = self.child.stdout.as_mut().ok_or(std::io::Error::new(
-            std::io::ErrorKind::BrokenPipe,
-            "can connect to child stdout",
-        ))?;
+        let child_stdout = self.child.stdout.as_mut().ok_or_else(|| {
+            std::io::Error::new(
+                std::io::ErrorKind::BrokenPipe,
+                "can connect to child stdout",
+            )
+        })?;
 
         let mut buffer = [0; 16];
-        child_stdout.read(&mut buffer[..])?;
+        child_stdout.read_exact(&mut buffer[..])?;
 
         let mut digits = String::new();
         let mut digit = [0; 1];
         loop {
-            child_stdout.read(&mut digit[..])?;
+            child_stdout.read_exact(&mut digit[..])?;
             let char_digit = digit[0] as char;
 
             if char_digit.is_digit(10) {
                 digits.push(char_digit);
             } else {
                 let mut whitespace = [0; 3];
-                child_stdout.read(&mut whitespace[..])?;
+                child_stdout.read_exact(&mut whitespace[..])?;
                 break;
             }
         }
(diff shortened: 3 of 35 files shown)


components/dada-execute/src/permission/my.rs
```diff
@@ -43,12 +43,14 @@ impl My {
 
     pub(super) fn check_read(&self, interpreter: &Interpreter) -> eyre::Result<()> {
         self.given.check_still_valid(interpreter)?;
-        Ok(self.tenant.cancel_tenant_if_exclusive(interpreter))
+        self.tenant.cancel_tenant_if_exclusive(interpreter);
+        Ok(())
```
nikomatsakis, line 47: Is this in response to a clippy warning? This feels like a regression to me -- I set it up that way explicitly, because I wanted a compilation error if the return type of `cancel_tenant_if_exclusive` changes from `()` to something else. Admittedly, it looks a bit weird and I don't always do this very consistently.

nikomatsakis: (Sometimes I think I'd prefer to have Rust warn if you ever don't consume a return value, full stop.)

components/dada-ir/src/span.rs
```diff
@@ -76,6 +76,7 @@ impl Span {
     }
 
     /// Returns a 0-length span at the start of this span
+    #[must_use]
```
nikomatsakis, line 79: ?

components/dada-lang/src/test_harness/lsp_client.rs
```diff
@@ -185,7 +185,7 @@ impl ChildSession {
 pub enum LspCommand {
     initialize {
         id: usize,
-        params: lsp_types::InitializeParams,
+        params: Box<lsp_types::InitializeParams>,
```
nikomatsakis, line 188: Is this from a clippy warning?

nikomatsakis: Because of uneven enum variant sizes?

nikomatsakis: If so, very cool!

components/dada-lex/src/lines.rs
```diff
@@ -36,6 +36,7 @@ pub fn line_column(db: &dyn crate::Db, filename: Filename, position: Offset) ->
 }
 
 #[salsa::memoized(in crate::Jar ref)]
+#[allow(clippy::needless_lifetimes)]
```
nikomatsakis, line 39: ...this is annoying, maybe I will tweak the salsa output to generate that lint, or to avoid it

components/dada-parse/src/parser.rs
```diff
@@ -210,10 +210,8 @@ trait ParseList {
             // List items can always be separated by a newline...
             if !self.skipped_newline() {
                 // ...otherwise, they *may* require a separator
-                if comma_separated {
-                    if !self.eat_comma() {
-                        break;
-                    }
+                if comma_separated && !self.eat_comma() {
+                    break;
```
nikomatsakis, line 214: oddly enough, I find this harder to read, I put the nesting there on purpose, but ... it's ok :)

nikomatsakis: (my brain just has trouble parsing `&&` things...)

components/dada-parse/src/parser/items.rs
```diff
@@ -29,10 +29,8 @@ impl<'db> Parser<'db> {
     fn parse_item(&mut self) -> Option<Item> {
         if let Some(class) = self.parse_class() {
             Some(Item::Class(class))
-        } else if let Some(function) = self.parse_function() {
-            Some(Item::Function(function))
         } else {
-            None
+            self.parse_function().map(Item::Function)
```
nikomatsakis, line 33: Did this come from clippy? I find this a regression. It's both harder to read and harder to add more cases in the future. I think I'd be inclined to disable whatever clippy lint this is (crate-wide).

nikomatsakis (changes requested):
This is great, thanks! I left one or two nits, but I'm generally in favor of adopting clippy, though I've actually never used it before. I'm surprised to see some things get reformatted, as i always use "format on save" -- do you know if there's any kind of workflow where, instead of erroring on bad formatting, it pushes a "reformat source" commit to the branch of the PR?

components/dada-execute/src/permission/my.rs
nikomatsakis: (No need to change anything here, just me grumbling.)

nikomatsakis: image

lqd:
> do you know if there's any kind of workflow where, instead of erroring on bad formatting, it pushes a "reformat source" commit to the branch of the PR?

I've seen people do this (on PRs and on the main branch) with [this `rustfmt` action](https://github.com/mbrobbel/rustfmt-check), and which was made by the author of this very PR!

components/dada-lex/src/lines.rs
mbrobbel: Yes I just dropped those `#[allow(...)]`s here to make Clippy happy. Fixing it in salsa is the way to go.

components/dada-ir/src/span.rs
mbrobbel: It's this one: https://rust-lang.github.io/rust-clippy/master/#return_self_not_must_use

components/dada-execute/src/permission/my.rs
mbrobbel: It's this one: https://rust-lang.github.io/rust-clippy/master/index.html#unit_arg

```shell
error: passing a unit value to a function
  --> components/dada-execute/src/permission/my.rs:46:9
   |
46 |         Ok(self.tenant.cancel_tenant_if_exclusive(interpreter))
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `-D clippy::unit-arg` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unit_arg
help: move the expression in front of the call and replace it with the unit literal `()`
   |
46 ~         self.tenant.cancel_tenant_if_exclusive(interpreter);
47 +         Ok(())
   |

error: passing a unit value to a function
  --> components/dada-execute/src/permission/my.rs:51:9
   |
51 |         Ok(self.tenant.cancel_tenant(interpreter))
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unit_arg
help: move the expression in front of the call and replace it with the unit literal `()`
   |
51 ~         self.tenant.cancel_tenant(interpreter);
52 +         Ok(())
```

components/dada-lang/src/test_harness/lsp_client.rs
mbrobbel: Yep: https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant

components/dada-parse/src/parser.rs
mbrobbel: https://rust-lang.github.io/rust-clippy/master/index.html#collapsible_if

components/dada-parse/src/parser/items.rs
mbrobbel: Yep: https://rust-lang.github.io/rust-clippy/master/index.html#manual_map

```shell
error: manual implementation of `Option::map`
  --> components/dada-parse/src/parser/items.rs:32:16
   |
32 |           } else if let Some(function) = self.parse_function() {
   |  ________________^
33 | |             Some(Item::Function(function))
34 | |         } else {
35 | |             None
36 | |         }
   | |_________^ help: try this: `{ self.parse_function().map(Item::Function) }`
   |
   = note: `-D clippy::manual-map` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_map
```

I agree that this does not help here. The `None` is not only returned because `self.parse_function()` returns `None`, but because all the above options were also `None`. I'll add an `#[allow(...)]` here but I do think this lint can be useful elsewhere.

components/dada-ir/src/span.rs
nikomatsakis: Interesting. OK, not opposed. :)

components/dada-parse/src/parser/items.rs
nikomatsakis: I'm of mixed minds. In general I find that it's often harder to read code that makes extensive use of higher-level helpers like `map`, so sometimes I intentionally write things in a simpler way. On the other hand, I prefer to have fewer customizations where possible, so let's add the "local allow" for now, that seems reasonable.

components/dada-execute/src/permission/my.rs
nikomatsakis: OK. Not sure I agree with this particular one, but I'm going to let it ride. :)

nikomatsakis:
> I've seen people do this (on PRs and on the main branch) with [this `rustfmt` action](https://github.com/mbrobbel/rustfmt-check), and which was made by the author of this very PR!

Got it.

nikomatsakis:
@mbrobbel so I tried to integrate rustfmt-check but I admit I wasn't 100% sure where would be the most appropriate place to insert it -- care to open a PR editing the `.yml` file?

Issue #14, opened 2022-01-07
Debug dumps for syntax tree


nikomatsakis:
The validated trees have pretty decent debug output:

https://github.com/dada-lang/dada/blob/d86c5f170b4504e7ffa0c7cd655b4577987e62f5/dada_tests/hello_world.validated#L1-L19

but syntax trees, not so much:

https://github.com/dada-lang/dada/blob/d86c5f170b4504e7ffa0c7cd655b4577987e62f5/dada_tests/hello_world.syntax#L1-L5

We need to write various `DebugWithDb` impls for syntax, much like this one:

https://github.com/dada-lang/dada/blob/d86c5f170b4504e7ffa0c7cd655b4577987e62f5/components/dada-ir/src/code/validated.rs#L189-L264

PR #15, opened 2022-01-08
Add `DebugWithDb` impls for some of the syntax nodes

lqd:
This PR is work towards #14, and adds some debug impls for the syntax nodes currently used in tests.

Surely could be improved further, if that's not what you had in mind.

### DIFF
--- a/components/dada-ir/src/token.rs
+++ b/components/dada-ir/src/token.rs
@@ -78,3 +78,20 @@ impl Token {
         }
     }
 }
+
+impl<Db: ?Sized + crate::Db> salsa::DebugWithDb<Db> for Token {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>, db: &Db) -> std::fmt::Result {
+        match self {
+            Token::Alphabetic(word) => f.debug_tuple("Alphabetic").field(&word.debug(db)).finish(),
+            Token::Number(word) => f.debug_tuple("Number").field(&word.debug(db)).finish(),
+            Token::Prefix(word) => f.debug_tuple("Prefix").field(&word.debug(db)).finish(),
+            Token::Tree(tree) => f.debug_tuple("Tree").field(&tree.debug(db)).finish(),
+            Token::FormatString(format_string) => f
+                .debug_tuple("FormatString")
+                .field(&format_string.debug(db))
+                .finish(),
+            Token::Comment(_) => write!(f, "Comment"),
+            _ => std::fmt::Debug::fmt(self, f),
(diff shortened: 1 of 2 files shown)


components/dada-ir/src/token_tree.rs
```diff
@@ -16,8 +11,11 @@ salsa::entity2! {
 impl<Db: ?Sized + crate::Db> salsa::DebugWithDb<Db> for TokenTree {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>, db: &Db) -> std::fmt::Result {
         let db = db.as_dyn_ir_db();
-        let file_span: FileSpan = self.span(db).in_file(self.filename(db));
-        write!(f, "Tokens({:?})", file_span.into_debug(db))
+        let file_span = self.span(db).in_file(self.filename(db));
+        f.debug_struct("TokenTree")
+            .field("span", &file_span.into_debug(db))
+            .field("tokens", &self.tokens(db).debug(db))
+            .finish()
```
nikomatsakis, line 18: I think I intentionally kept this debug-with minimal, to avoid spamming the output

components/dada-ir/src/token.rs
```diff
@@ -78,3 +78,20 @@ impl Token {
...
+            Token::Prefix(word) => f.debug_tuple("Prefix").field(&word.debug(db)).finish(),
+            Token::Tree(tree) => f.debug_tuple("Tree").field(&tree.debug(db)).finish(),
+            Token::FormatString(format_string) => f
+                .debug_tuple("FormatString")
+                .field(&format_string.debug(db))
+                .finish(),
+            Token::Comment(_) => write!(f, "Comment"),
+            _ => std::fmt::Debug::fmt(self, f),
+        }
+    }
```
nikomatsakis, line 96: But this seems useful

components/dada-ir/src/token_tree.rs
nikomatsakis: Maybe revert this change?

nikomatsakis (changes requested):

components/dada-ir/src/token_tree.rs
lqd: Ah I see, I can do that. Since that will turn the debug dump back to the starting point mentioned in issue #14 I misunderstood that it would be an improvement. I'll push a commit to remove this change, and look at the other syntax nodes.

nikomatsakis (approved):
this seems useful on its own, going to land :)

PR #16, opened 2022-01-08
Implement classes

nikomatsakis:
This makes classes execute and more-or-less work, which required fixes in the parser, validator, and elsewhere. 

Along the way, I tweaked how logging works. You can use `DADA_LOG` to configure it, but you can also do `cargo run -- --log ....`. Not sure about the whole `DADA_LOG` thing, maybe that should be `RUST_LOG`...? Don't have a strong opinion.


nikomatsakis:
I'm not going to bother to ask for review, but I figured I would get in the habit of starting to open PRs. :stuck_out_tongue:

PR #26, opened 2022-01-11
Error when label does not match

XFFXFF:
https://github.com/dada-lang/dada/blob/3ebc34f5dd954349238d740aa288d3ac3a2516e9/components/dada-execute/src/data.rs#L212-L225
We should return error when `expected_name != actual _word`, not when they are equal.


nikomatsakis (approved):

nikomatsakis:
Nice, how did I miss that? :laughing:

PR #30, opened 2022-01-11
A few parser additions

brson:
I really wanted to be able to break out of loops and discovered none of it was implemented. When I got to the interpreter for ops and found `todo!` I gave up for now. I'll probably wait until you sketch out how an op is interpreted.

In the meantime here is parsing for `loop`, `true`, `false`, `==`, `<`, and `>`. I just followed the patterns in the parser to hopefully get the precedence right for the ops.

### DIFF
--- a/components/dada-parse/src/parser/code.rs
+++ b/components/dada-parse/src/parser/code.rs
@@ -173,14 +173,31 @@ impl CodeParser<'_, '_> {
             return Some(expr);
         }
 
-        self.parse_expr_3()
+        self.parse_expr_4()
+    }
+
+    pub(crate) fn parse_expr_4(&mut self) -> Option<Expr> {
+        let mut expr = self.parse_expr_3()?;
+
+        loop {
+            if let Some(expr1) = self.parse_binop(expr, &[Op::Plus, Op::Minus], Self::parse_expr_3)
+            {
+                expr = expr1;
+                continue;
+            }
+
+            break;
+        }
+
+        Some(expr)
     }
 
     pub(crate) fn parse_expr_3(&mut self) -> Option<Expr> {
         let mut expr = self.parse_expr_2()?;
 
         loop {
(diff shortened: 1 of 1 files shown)


components/dada-parse/src/parser/code.rs
```diff
@@ -196,9 +213,11 @@ impl CodeParser<'_, '_> {
...
 
         loop {
-            if let Some(expr1) =
-                self.parse_binop(expr, &[Op::DividedBy, Op::Times], Self::parse_expr_1)
-            {
+            if let Some(expr1) = self.parse_binop(
+                expr,
+                &[Op::EqualEqual, Op::LessThan, Op::GreaterThan],
+                Self::parse_expr_1,
+            ) {
```
nikomatsakis, line 220: I think the precedence here isn't quite right. I guess the numbering is a bit counterintuitive, now that I think about it, but the idea is that `parse_expr_3` is *lower* precedence than `parse_expr_2`. Maybe we should rename them to count the other way. I think I just wanted to be able to add "new layers" without renumbering anything.

We want to be able to do `x == y * z` and have that parse as `x == (y * z)`, I imagine; for that to work, you want:

* expr 4 => `==`, `<=`, `>=`, etc
* expr 3 => `+`, `-`
* expr 2 => `/`, `*`

nikomatsakis (changes requested):

components/dada-parse/src/parser/code.rs
brson: Ah I just goofed. The numbering is fine. I was even staring at the rust operator precedence table while I wrote it, and still put them in the wrong order.

brson: I think it's correct now.

nikomatsakis (approved):

nikomatsakis:
There is no test, but that's ok, it would probably fail to get through later phases

Issue #28, opened 2022-01-11
Add a runtime exception to the license


brson:
It's not pressing now, but at some point people may use dada, and it would be nice to avoid the mistake Rust made of not having a runtime exception in its license.

A runtime exception allows people to distribute binaries of software that contains runtime code like the standard library without any attribution, and is common for most open source standard libraries.

Rust doesn't have this exception and it's a quietly huge mess that will probably never be fixed.

It would be nice to set the right precedent for the dada ecosystem and choose a licensing regime with a runtime exception.

It doesn't matter yet since there's no runtime or standard library and only a few contributors, but the more contributors the harder it is to relicense things.

A simple interim solution would be to add a CC0 dual license, which is effectively public domain, and gives maximum freedom to relicense more perfectly in the future.

nikomatsakis:
Seems reasonable to me. We should probably adopt Apache 2 as well.

nikomatsakis:
If you have suggested wording changes, feel free to open a PR.

brson:
Well I suggested a bold CC0-based solution in https://github.com/dada-lang/dada/pull/38. I think it's the most general and leaves the most options open.

Others I've considered:

- MIT or Apache 2 + LLVM exception, though I have some concerns about the exact terms of the LLVM exception
- MIT or Apache 2 or Boost (or Zlib)

It's such a hard thing to get perfectly right, and as we've seen with Rust, the practical impact of getting it wrong is almost none, and very few people actually care.

brson:
Stale.

PR #38, opened 2022-01-12
Add Apache and CC0 licenses

brson:
This is a suggested temporary solution to https://github.com/dada-lang/dada/issues/28, which is concerned with "runtime exceptions".

The CC0 license effectively makes all contents of this repository public domain, relinquishing all copyright claims wherever legally possible.

The intent here is that the CC0 license makes it possible to relicense the codebase in the future, once somebody has made more concrete decisions about how to apply a runtime exception, without getting consent from all prior contributors, and without requiring a contributor license agreement (CLA). At that stage, the CC0 license can be removed if desired (and in the meantime the CC0 is an effective runtime exception), leaving prior code without copyright claims, but future code under the new license.

It does require contributors to completely relinquish any copyright claims to their contributions, but not to another entity as with a CLA.

I feel nervous about this as it is rare to completely give up all rights like this (though some like @burntsushi do it), but I also think the practical impact of downgrading from MIT to CC0 is effectively none, and it's a bold statement.

I am completely open to not doing this - alternate solutions can be discussed on the associated issue.

The Apache-2 license remains necessary for its clauses preventing contributors from using patents maliciously. The MIT license then is kind of superfluous with CC0, but is common and understandable and the MIT/Apache combo is standard in the Rust community.

As far as existing contributors are concerned, I think it is easy to argue that nobody but @nikomatsakis actually has copyrightable material here yet, so _this_ relicense can be done without explicit consent, but existing contributors should still approve of such a bold license choice cc @XFFXFF @lqd @mbrobbel.


nikomatsakis:
In principle, I love this. Let me dig a bit into the legal implications though.

nikomatsakis:
One question I have -- if we have deps that are MIT/Apache2, does this present an issue?

nikomatsakis:
On further reflection, feeling pretty good about this formulation.

brson:
> One question I have -- if we have deps that are MIT/Apache2, does this present an issue?

It doesn't present an issue to applying CC0 to this repo, but it does when it comes time to think about which parts of code truly need runtime exceptions. Just like in the Rust standard library we have to consider that anything in the library is MIT-equivalent, we would need to be careful that any code that might be compiled into a dada program has a runtime exception.

This would become an issue with Rust crates that are compiled into the dada runtime.

A related and distant concern is that if we establish a runtime exception for dada itself, then, like in the Rust ecosystem, we might hope that most other crates would pick the same license. This would solve a common problem in pretty much every package-based ecosystem where most projects distributing binaries ignore the attribution requirement for binaries, and in practice nobody really cares, but technically they are violating the license.

brson:
Having this repo be CC0 would also need we need to be careful not to just copy-paste code from other Rust projects.

brson:
One question I have is this - if CC0 says "I give up all copyright claims", but we're also distributing under license that indicate the code is copyrighted, what is the actual legal impact to the MIT and Apache license. It may be that the MIT license is completely invalidated in a way that doesn't really matter, but do the patent-protection clauses of Apache still hold?

nikomatsakis:
@brson Assuming we did *not* take this route...

> It doesn't present an issue to applying CC0 to this repo, but it does when it comes time to think about which parts of code truly need runtime exceptions. Just like in the Rust standard library we have to consider that anything in the library is MIT-equivalent, we would need to be careful that any code that might be compiled into a dada program has a runtime exception.

...it seems like any deviation from the Rust license is going to cause a potential issue here, right? i.e., the structure of our license is going to be "MIT + Apache2 + X" where X is the "runtime exception". But that is already distinct from the license of most every Rust library we use, which is MIT + Apache2.

brson:
> @brson Assuming we did _not_ take this route...
> 
> > It doesn't present an issue to applying CC0 to this repo, but it does when it comes time to think about which parts of code truly need runtime exceptions. Just like in the Rust standard library we have to consider that anything in the library is MIT-equivalent, we would need to be careful that any code that might be compiled into a dada program has a runtime exception.
> 
> ...it seems like any deviation from the Rust license is going to cause a potential issue here, right? i.e., the structure of our license is going to be "MIT + Apache2 + X" where X is the "runtime exception". But that is already distinct from the license of most every Rust library we use, which is MIT + Apache2.
> 
> But I guess that the narrower we can make X, the better.

Yes. Since a large amount of the code will be 3rd-party Rust crates, dada will mostly by Rust-licensed. The only hope for a runtime exception making a meaningful difference is if we can create a boundary around the dada runtime through which we can keep Rust-licensed code out, or at least limited, with a pathway to replacing it.

To be clear this concern doesn't apply to the interpreter, for which there's no hope of losing the Rust license, but for code directly compiled into compiled dada distributables.

brson:
That also includes things like a dada standard library. So having a runtime exception could be more impactful for a prospective dada ecosystem, even if the ship has sailed for the rust ecosystem.

brson:
With the dada runtime likely being written largely in Rust, it will be difficult to completely get rid of the Rust license, though I could schemes to limit how much of Rust core/std is used, with _some_ plan to eliminate it.

nikomatsakis:
Well, this has sat here for a while! I'm not really sure what I'm waiting for. I like the idea of this PR, but I also feel like it may be impractical to really do anything but conform to the general Rust license setup, imperfect as it is.

nikomatsakis:
I'm going to close this for now, but we should revisit the question at some point.

PR #45, opened 2022-01-13
A number of refinements to dada-web

### LINKED ISSUE #41: better formatting for the dada playground
The dada playground is "up and going" at https://dada-lang.org/playground/, but it is not very attractive. It would probably be better if it looked more like https://play.rust-lang.org/ -- i.e., formatted as two rows, or at least width dependent, etc.

### LINKED ISSUE #43: must edit for dada playground code to run
The dada playground requires you to edit a character before the output is updated. This is silly.

brson:
A lot of this is to be responsive to mobile screen sizes. I also increased the font size a bit and made minor tweaks to the appearance. The output is also displayed on initial load now, which I think helps explain the lack of a "run" button.

Fixes #43 
Fixes #41

### DIFF
--- a/components/dada-web/index.css
+++ b/components/dada-web/index.css
@@ -2,6 +2,7 @@ body {
     display: flex;
     flex-wrap: wrap;
     min-height: 800px;
+    font-size: 18px;
 }
 
 .area {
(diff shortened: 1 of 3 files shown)


brson:
Fixes https://github.com/dada-lang/dada/issues/43

nikomatsakis (approved):
Much nicer =)

PR #47, opened 2022-01-13
Make dada the default bin to run

brson:
With the addition of `xtask` `cargo run` no longer runs dada without further flags. It seems like most people will want `run` to just run dada. What do you think?


nikomatsakis:
Agreed, I just didn't know about that feature. =)

PR #55, opened 2022-01-14
quote Duchamp for the initial playground program

nikomatsakis:
This is more fun.


lqd:
I love these playground previews in PRs ! (and have long wished we could add such rustc PR previews on the rust playground :)

Issue #8, opened 2022-01-06
operators are not implemented


nikomatsakis:
We parse operators but they're not really implemented correctly. How I expect them to work:

* [ ] move `dada_ir::op` into `dada_ir::code::syntax::op`
* [ ] introduce `dada_ir::code::validated::op` which contains *only* the "unary/binary operators" that we want in the end -- i.e., `+=` is not a validated op, but `+` is. That sort of thing.
* [ ] make `dada_ir::code::bir` take a `validated::op::Op`; maybe make it not hardcoded to binary operators, now that I think about it?
* [ ] in the interpreter, execute lhs to a value, then rhs, then apply the op.

I'm *tempted* to remove ops and replace them with intrinsics, but it seems like it'd be nice to not have every operator end the basic block. This might argue in favor of extended basic blocks in BIR but I don't want to think that hard.

brson:
@nikomatsakis I see you removed the help wanted label here. Are you going to do this yourself?

nikomatsakis:
No, I don't know why I removed that really.

brson:
I'm giving this a try.

brson:
I have this implemented for obvious cases here: https://github.com/brson/dada/tree/ops

I am going to take another day to write negative test cases and think about a few more cases.

brson:
At the moment there is no way to create signed integers, so that code cannot be tested.

nikomatsakis:
@brson feel free to open a draft PR, that would make it easier to review.

nikomatsakis:
I think I was trying to figure out the difference between "help wanted" and "good first issue".

nikomatsakis:
And decided "help wanted" would be like "Distinct skill sets from ones that Niko possesses"

PR #56, opened 2022-01-14
Interpret binops

brson:
This is wip for https://github.com/dada-lang/dada/issues/8. It is mergeable as is but there are a number of obvious things left do do, such as:

- write negative tests for bogus type / op combos
- do something on divide by zero. what?
- put todo!s on ops that I know should be implemented that I don't intend to think about yet, like float eq

There are also some seemingly missing ops like negation, not-equals, less-than-or-equals, greater-than-or-equals. And I can't find a way to create signed ints, so no tests for them.

I'm happy to restructure the code any way you like.


brson:
I wonder how thorough you want the test suite to be. I've now added 5 distinct tests each for a single operator that doesn't work with bools. There's lots more I can add like that. Maybe you don't want to deal with so many minor test cases yet.

dada_tests/interpret/ops_uint.dada
```diff
@@ -0,0 +1,66 @@
+async fn main() {
+    var uint_max = 18446744073709551615
+
+    if 1 + 1 == 2 {
+        print("1 + 1 == 2").await
+    }
+
+    # wrapping overflow
+    if uint_max + 2 == 1 {
```
nikomatsakis: we should decide what to do here =) bah, the annoying things. Maybe return an Err for now?

nikomatsakis (approved):
This looks good to me, I'd be up to land it as is, though we should open a FIXME to decide what to do about overflow.

nikomatsakis:
> do something on divide by zero. what?

I think overflow and divide by zero should yield an error for now (i.e., return `Err` for the `eyre::Result<Value>`)

nikomatsakis:
> I wonder how thorough you want the test suite to be. I've now added 5 distinct tests each for a single operator that doesn't work with bools. There's lots more I can add like that. Maybe you don't want to deal with so many minor test cases yet.

No strong opinion, I guess. =) I am happy to have more tests.

nikomatsakis:
> And I can't find a way to create signed ints, so no tests for them.

Yeah I was too lazy to support these. =) I guess we should add some kind of "suffixes" to the parser (we shouldn't need new tokens for that, we can use the "skipped whitespace" flag) for the time being.

brson:
@nikomatsakis I've changed oveflow to errors, divide by zero to errors, added a special case error for signed division overflow and added negative tests for bad string and unit binops.

Other types like floats are left for the future. I'll file a few issues.

nikomatsakis (approved):

nikomatsakis:
bors r+

Cool =)

PR #59, opened 2022-01-15
Download deploy

### LINKED ISSUE #50: xtask depoy takes a long time
I love the way I setup xtask deploy because it's self-contained: if you've got Rust installed, it works. I don't like configuring CI so anything I can do to make it simple is a win.

OTOH, I hate the way I setup xtask deploy because it takes ages to build, and netlify deployment is slow. 

Maybe we should make it download binaries for wasm-pack and mdbook instead? 

I also *feel* like cargo is building way more than it has to in order to run xtask, but I could be wrong. Looking at cargo-tree I saw an awful lot of stuff in there.

nikomatsakis:
The deploy command now downloads instead of building from scratch. It hardcodes x86_64....maybe should fix that.

Fixes #50

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,21 +38,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ammonia"
-version = "3.1.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


nikomatsakis:
Deploy time went from 13min per build to 3 =)

PR #65, opened 2022-01-16
strip whitespace when it uses tabs

### LINKED ISSUE #61: strip whitespace when it uses tabs, too
#57 removed leading whitespace (see #51), but it only handles spaces, not tabs. See this comment:

This is probably good enough, but it only strips leading *spaces*.

I feel like 

```
\t\tFoo
\t\tBar
```

should also work. To make it work, we would want to do

```rust
take_while(|c| c.is_whitespace())
```

and then collect into a *String*, I think. Not as simple, though.

_Originally posted by @nikomatsakis in https://github.com/dada-lang/dada/pull/57#discussion_r785300730_

XFFXFF:
This pr attemps to fix #61 . 
I don't know if there is a better way to do this, feel free to give suggestions.

### DIFF
--- a/components/dada-validate/src/validate/validator.rs
+++ b/components/dada-validate/src/validate/validator.rs
@@ -564,7 +564,7 @@ fn convert_to_dada_string(s: &str) -> String {
 
     if common_indentation != 0 {
         let first_prefix = whitespace_prefix_iter.next().unwrap_or_default();
-        while let Some(prefix) = whitespace_prefix_iter.next() {
+        for prefix in whitespace_prefix_iter {
             let common_prefix_len = first_prefix[..common_indentation]
                 .as_bytes()
                 .iter()
--- a/components/dada-validate/src/validate/validator.rs
+++ b/components/dada-validate/src/validate/validator.rs
@@ -14,7 +14,6 @@ use dada_ir::storage_mode::StorageMode;
 use dada_ir::word::Word;
 use dada_lex::prelude::*;
 use dada_parse::prelude::*;
-use std::cmp::min;
 use std::rc::Rc;
 use std::str::FromStr;
 
@@ -541,6 +540,10 @@ impl<'me> Validator<'me> {
     }
 }
 
+fn count_bytes_in_common(s1: &[u8], s2: &[u8]) -> usize {
+    s1.iter().zip(s2).take_while(|(c1, c2)| c1 == c2).count()
+}
+
 // Remove leading, trailing whitespace and common indent from multiline strings.
 fn convert_to_dada_string(s: &str) -> String {
     // If the string has only one line, leave it and return immediately.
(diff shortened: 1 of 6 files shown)


XFFXFF:
I didn't add a test like
```
\t\tFoo
\t\tBar
```
Because "\t" will be lexed as "\\\t" now, which is not a whitespace.

components/dada-validate/src/validate/validator.rs
```diff
@@ -540,22 +541,49 @@ impl<'me> Validator<'me> {
...
-        .lines()
-        .filter(|&line| !line.trim().is_empty())
-        .collect::<Vec<&str>>();
-
-    let indent = non_empty_lines
-        .iter()
-        .map(|line| line.chars().into_iter().take_while(|c| *c == ' ').count())
+
+    // Split string into lines and filter out empty lines.
+    let non_empty_line_iter = s.lines().filter(|&line| !line.trim().is_empty());
```
nikomatsakis, line 552: A thought--

The common prefix has to be the prefix of *all* the lines. So maybe start with the first line and find its whitespace prefix. Then iterate over all the other lines. For each line, you compute the number of characters the two have in common and you find the minimum N of that. Then you keep the first N characters of the prefix.

In pseudocode...

```
fn count_characters_in_common(s1: &str, s2: &str) -> usize {
    s1.chars().zip(s2.chars()).take_while(|(c1, c2)| c1 == c2).count()
}

let non_empty_line_iter = s.lines().filter(|&line| !line.trim().is_empty());
let first_line = non_empty_line_iter.next();
let mut prefix = whitespace_prefix_of(first_line);
let shared_prefix_chars = non_empty_line_iter.map(|s| count_characters_in_common(prefix, s)).min();
prefix = &prefix[..shared_prefix_chars];
strip_prefix_from_original_string(prefix);
```

nikomatsakis (changes requested):

components/dada-validate/src/validate/validator.rs
XFFXFF: This will make the code simpler and more efficient, thanks! I'll refactor it latter.

nikomatsakis: One note is that the code I wrote counts characters, but for slicing, you really want to count utf-8 width. You can do that in various ways -- e.g. [`len_utf8`](https://doc.rust-lang.org/std/primitive.char.html#method.len_utf8) or the iterator (forgot the name) that gives you both characters and byte offsets.

XFFXFF: I modified the code to count characters to count bytes, I think this should be equivalent to `len_utf8`.

components/dada-validate/src/validate/validator.rs
```diff
@@ -540,32 +540,47 @@ impl<'me> Validator<'me> {
...
+    if let Some(first_line) = non_empty_line_iter.next() {
+        let prefix = first_line
+            .chars()
+            .into_iter()
+            .take_while(|c| c.is_whitespace())
+            .collect::<String>();
+        let common_indent = non_empty_line_iter
+            .map(|s| count_bytes_in_common(prefix.as_bytes(), s.as_bytes()))
+            .min()
+            .unwrap_or(0);
```
nikomatsakis, line 566: Probably fine, yes. There might be some weird edge case in unicode I don't know about but I guess we can take our chances. =)

nikomatsakis (approved):

PR #74, opened 2022-01-17
Add support for some escapes

### LINKED ISSUE #66: Support escapes in our strings.
Our strings do not support escapes like `\t`, `\n`.
`print("Foo \nBar")` now gives 
```
Foo \nBar
```
it should give
```
Foo
Bar
```

XFFXFF:
fix #66


components/dada-lex/src/lex.rs
```diff
@@ -203,6 +215,19 @@ where
                 break;
             }
 
+            if ch == '\\' {
```
nikomatsakis, line 218: I think this should be done in a later phase, after lexing-- in the parser or during validation, for example. In particular, I think that explicitly written `\n` and `\t` etc should never be stripped by the whitespace normalization code.

XFFXFF: Got it, I'll fix it later.

dada_tests/interpret/string-escape.dada
```diff
@@ -0,0 +1,21 @@
+async fn main() {
+    print("Foo\nBar").await
+
+    print("Foo\tBar").await
+
+    print("Foo\r\nBar").await
+
+    print("Foo \\ Bar").await
+
+    print("Foo \" Bar").await
```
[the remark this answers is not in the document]
XFFXFF: Should we trim the explicitly written `\n` and `\t` at the start of a line?
```
print("
\tFoo
\tBar
")
```
should give
```
Foo
Bar
```
or
```
    Foo
    Bar
```

components/dada-lex/src/lex.rs
XFFXFF: I modified it, feel free to request changes.

XFFXFF:
I removed the test case `"Foo \" Bar"`， because the lexer now thinks `"Foo \"` is a string literal. That's another bug, I'll open an issue later.
https://github.com/dada-lang/dada/blob/ecd3aab2a960d689b22c65b00d4009cc26611791/components/dada-lex/src/lex.rs#L197-L205

dada_tests/interpret/string-escape.dada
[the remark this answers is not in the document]
nikomatsakis: I think it should give `\tFoo\tBar` -- i.e., we trim the newlines, but we keep the stuff written with `\t`

nikomatsakis (approved):

PR #75, opened 2022-01-17
Print the token tree in its DebugWithDb

brson:
While hacking on other things I found that TokenTree's were not debug-printing in a useful way for my purposes.

cc https://github.com/dada-lang/dada/issues/14

### DIFF
--- a/components/dada-ir/src/token_tree.rs
+++ b/components/dada-ir/src/token_tree.rs
@@ -1,9 +1,4 @@
-use crate::{
-    filename::Filename,
-    span::{FileSpan, Span},
-    token::Token,
-    Jar,
-};
+use crate::{filename::Filename, span::Span, token::Token, Jar};
 
 salsa::entity2! {
     entity TokenTree in Jar {
@@ -16,8 +11,11 @@ salsa::entity2! {
 impl<Db: ?Sized + crate::Db> salsa::DebugWithDb<Db> for TokenTree {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>, db: &Db) -> std::fmt::Result {
         let db = db.as_dyn_ir_db();
-        let file_span: FileSpan = self.span(db).in_file(self.filename(db));
-        write!(f, "Tokens({:?})", file_span.into_debug(db))
+        if !f.alternate() {
+            write!(f, "{:?}", self.tokens(db))
+        } else {
+            write!(f, "{:#?}", self.tokens(db))
+        }
     }
 }
 
--- a/dada_tests/class/class-field-0/syntax.ref
+++ b/dada_tests/class/class-field-0/syntax.ref
@@ -2,9 +2,109 @@
     syntax::Tree {
         origin: Code(
             Some(
-                Tokens(dada_tests/class/class-field-0.dada:3:15:3:15),
+                [],
             ),
-            Tokens(dada_tests/class/class-field-0.dada:3:18:6:1),
+            [
+                Whitespace(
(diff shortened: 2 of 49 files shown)


dada_tests/class/class-field-0/syntax.ref
```diff
@@ -2,9 +2,109 @@
...
+                Whitespace(
+                    ' ',
+                ),
+                Whitespace(
+                    ' ',
+                ),
+                Alphabetic(
+                    Word(
+                        Id {
+                            value: 7,
```
XFFXFF, line 26: Should we expand word to string?

brson:
After a few more hours of experience I'm not sure this is quite the right thing to do. It doesn't match the pattern used in other syntax entities, like `Tree`, where _some_ fields of `Tree` are printed, but not all.

But I am also finding that `Tree` doesn't print what I expect either - it is just printing the tokens, but not the parsed expression.

I'd like to take another pass at this yet, but I'll leave it here for comments.

XFFXFF:
I think something like this might be helpful :)
image

components/dada-ir/src/token_tree.rs
```diff
@@ -16,8 +11,11 @@ salsa::entity2! {
 impl<Db: ?Sized + crate::Db> salsa::DebugWithDb<Db> for TokenTree {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>, db: &Db) -> std::fmt::Result {
         let db = db.as_dyn_ir_db();
-        let file_span: FileSpan = self.span(db).in_file(self.filename(db));
-        write!(f, "Tokens({:?})", file_span.into_debug(db))
+        if !f.alternate() {
+            write!(f, "{:?}", self.tokens(db))
+        } else {
+            write!(f, "{:#?}", self.tokens(db))
+        }
```
nikomatsakis, line 18: So, I rather intentionally kept token trees from printing all the tokens, because the output is overwhelming. But I do think it's useful to have this output. I'm not sure the best way to balance this, but clearly we need a comment either way.

I *think* I would prefer that you "opt-in" to printing the tokens by e.g. doing `token_tree.details`, but I wonder if what we actually want is some configuration, like rustc's `-Zverbose`?

components/dada-ir/src/token_tree.rs
```diff
@@ -16,8 +11,11 @@ salsa::entity2! {
 impl<Db: ?Sized + crate::Db> salsa::DebugWithDb<Db> for TokenTree {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>, db: &Db) -> std::fmt::Result {
         let db = db.as_dyn_ir_db();
-        let file_span: FileSpan = self.span(db).in_file(self.filename(db));
-        write!(f, "Tokens({:?})", file_span.into_debug(db))
+        if !f.alternate() {
+            write!(f, "{:?}", self.tokens(db))
```
nikomatsakis, line 15: Try this:

```suggestion
            write!(f, "{:?}", self.tokens(db).debug(db))
```

components/dada-ir/src/token_tree.rs
brson: I was wondering if that was the case.

brson:
This patch is not correct in several dimensions and I'm not immediately planning to fix it so I'm closing it. Per niko's previous comment, we need someway to see this info, but the standard debug view isn't it because it produces huge output and explodes the test cases.

In the meantime, with this specific case I can just reach into the tree and print the tokens themselves.

PR #77, opened 2022-01-17
`\"` is not the terminator of a string

### LINKED ISSUE #76: `\"` is not the terminator of a string
For example `"Foo \" Bar"` , Now the lexer thinks `"Foo \"` is a string literal
https://github.com/dada-lang/dada/blob/ecd3aab2a960d689b22c65b00d4009cc26611791/components/dada-lex/src/lex.rs#L197-L205

XFFXFF:
fix #76


XFFXFF:
Oh! I forgot the case `\\"`

PR #83, opened 2022-01-19
parse `+=` and friends

### LINKED ISSUE #69: Tests for `+=` and friends
@brson added binop support in #56 but I don't *think* it included tests for `+=`. We should add a test like

```
async fn main() {
    var x = 22
    x += 44
    if x == 66 {
        print("Huzzah").await
    } else {
        print("Uh-oh").await
    }
}
```

XFFXFF:
fix #69


nikomatsakis (changes requested):
Cool! Can we add a test for what happens if you do

```rust
foo() += 3
```

I think that ought to be a validation error, as `foo()` is a value, not a place in memory, but I suspect it is not right now. We could fix that in a follow-up PR (in which case I'd want to file an issue).

XFFXFF:
image
As you think, there is no validation error here.  
I'm not sure if you want me to add the test to the dada repo or you just want to check the results 😂

nikomatsakis:
@XFFXFF I'd say create a test and open an issue, and then you can r=me this PR

XFFXFF:
bors r=me

PR #85, opened 2022-01-20
visualize the heap graph in the playground

### LINKED ISSUE #14: Debug dumps for syntax tree

### LINKED ISSUE #52: spans need to remove whitespace
The spans produced by parser often include the whitespace before/after the relevant tokens, resulting in strange errors:

image

I think I would modify the parser so that when it is attempt to create a syntax node with a span, it first does a "trim" step to cut whitespace from either side?

nikomatsakis:
This PR allows people to click in the middle of the source in the playground and see the current state of the heap graph. It's still a WIP but it kind of works.

Fixes #52 
Fixes #14


brson:
Fun.

The biggest problem I'm seeing is that sometimes wasm panicks with an index out of bounds and then everything quietly stops working. This happens when I put the courser at the end of the source. Would be nice to fix that at least before deploying it.

nikomatsakis:
@brson yeah, I fixed that problem

PR #89, opened 2022-01-22
cleanup wasm integration

nikomatsakis:
We now have tracing logs, panic hooks, and a more extensible API.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -135,6 +135,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "console_error_panic_hook"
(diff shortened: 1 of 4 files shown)


nikomatsakis:
r? @brson -- I think you've played around more with wasm than I have, any ideas for other common setup hooks?

It'd be nice if the log level was more configurable: ideally I'd like to be able to put in a precise environment string. But I couldn't quite figure out how to configure tracing that well, so I left it at this for now.

nikomatsakis:
Too late! Closing in favor of #90.

PR #96, opened 2022-01-25
permits specifying the start and end column of a diagnostic

### LINKED ISSUE #84: test harness doesn't permit specifying the end column of a diagnostic
I left some code in the test harness to specify the columns of a diagnostic but never modified the "expected diagnostics" function to fill that field in. I expected to parse something like this

```
fn foo() {
    blah = 3
    blah += baz
    #!      ^^^ ERROR unknown identifier
}
```

and then have the system check that the highlight was as expected.

To make this work, the first thing is to modify the `expected_diagnostics` function:

https://github.com/dada-lang/dada/blob/6dcc38a60546b499d862ced9d46f8c7fc6996ac7/components/dada-lang/src/test_harness.rs#L451

so that the regular expression can match the above and then set these fields to something other than `None`:

https://github.com/dada-lang/dada/blob/6dcc38a60546b499d862ced9d46f8c7fc6996ac7/components/dada-lang/src/test_harness.rs#L472-L473

After that, the rest of the code might need modification too, not sure.

XFFXFF:
This pr tries to fix #84, but it does not work when you want to highlight across lines.


dada_tests/validate/unknown_identifier.dada
```diff
@@ -0,0 +1,5 @@
+fn foo() {
```
XFFXFF, line 1: I add a test here.

nikomatsakis:
Cool =)

components/dada-lang/src/test_harness.rs
```diff
@@ -516,8 +516,10 @@ enum QueryKind {
 fn expected_diagnostics(path: &Path) -> eyre::Result<Vec<ExpectedDiagnostic>> {
     let file_contents = std::fs::read_to_string(path)?;
 
-    let re =
-        regex::Regex::new(r"^(?P<prefix>[^#]*)#!\s*(?P<severity>[^\s*]+)\s*(?P<msg>.*)").unwrap();
+    let re = regex::Regex::new(
+        r"^(?P<prefix>[^#]*)#!\s*(?P<highlight>\^+)?\s*(?P<severity>[^\s*]+)\s*(?P<msg>.*)",
```
nikomatsakis, line 520: I think, @XFFXFF, that we are going to want to have a few options here, kind of like with the `#?` annotations. I think it should probably be:

* `#! ERROR` -- error that starts on this line, useful for multi-line messages
* `#! ^^^^ ERROR` -- error with complete span information

XFFXFF: @nikomatsakis I'm not sure I understand what you meant, actually, both `#! ERROR` and `#! ^^^^ ERROR` are already  supported, `(?P<highlight>\^+)?` will optionally capture `^`, for `#! ERROR`, `c.name("highlight")` will be `None`

nikomatsakis: ah, great, I didn't read that closely, clearly

PR #97, opened 2022-01-26
Improve highlighting, show before/after

nikomatsakis:
The idea here is to help in tracking what expression the cursor is on and visualizing the state. But my CSS skills aren't really up to making the layout of this look remotely good. I figured I'd open it for now.


nikomatsakis:
I'm going to merge this for now.

bors r+

It hasn't reached a perfect end point but I've realized I want to try a different thing.

PR #100, opened 2022-02-01
Make checking for runtime diagnostics the same as compiletime

### LINKED ISSUE #6: dada files in test harness should have an annotation for errors at interpreter time
See the test:

https://github.com/dada-lang/dada/blob/cff3102697ef4dee168db7e53da3a68960224a29/dada_tests/interpret/print_wrong_label.dada#L1-L3

It generates an error:

https://github.com/dada-lang/dada/blob/cff3102697ef4dee168db7e53da3a68960224a29/dada_tests/interpret/print_wrong_label.stdout#L1

But the dada file doesn't have any `#` comment indicating that error is expected. This seems inconsistent with the way we handle errors during parsing. I'd prefer if errors generated at runtime had to be commented (and no comment meant "runs successfully").

cc @brson -- this seems like one you might like

brson:
Fixes #6

A side effect here is that an error diagnostic returned by the interpreter is no longer considered part of the standard output. This patch can in theory support multiple runtime diagnostics within the test harness, but the code to support that doesn't exist in the interpreter.


components/dada-lang/src/test_harness.rs
```diff
@@ -512,17 +536,25 @@ enum QueryKind {
...
+}
+
 /// Returns the diagnostics that we expect to see in the file, sorted by line number.
-fn expected_diagnostics(path: &Path) -> eyre::Result<Vec<ExpectedDiagnostic>> {
+fn expected_diagnostics(path: &Path) -> eyre::Result<ExpectedDiagnostics> {
     let file_contents = std::fs::read_to_string(path)?;
 
     let re = regex::Regex::new(
-        r"^(?P<prefix>[^#]*)#!\s*(?P<highlight>\^+)?\s*(?P<severity>ERROR|WARNING|INFO)\s*(?P<msg>.*)",
+        r"^(?P<prefix>[^#]*)#!\s*(?P<highlight>\^+)?\s*(?P<type>RUN)?\s*(?P<severity>ERROR|WARNING|INFO)\s*(?P<msg>.*)",
```
nikomatsakis, line 551: +1 this is cool

dada_tests/interpret/class-expected-1-found-2-unlabeled.dada
```diff
@@ -2,4 +2,5 @@ class Foo(arg)
 
 async fn main() {
     x = Foo(10, 22)
+    #!  ^^^^^^^^^^^ RUN ERROR expected to find 1 arguments, but found 2
```
nikomatsakis, line 5: Ah, I love the combo with exact spans. Nifty. :)

nikomatsakis (approved):
This is nice. I've been annoyed in my permission work that I can't annotate the result of the test that I expect, since most of those turn out to be runtime failures.

PR #102, opened 2022-02-02
Welcome to the machine

nikomatsakis:
This is the new interpreter / permission model I am working on. It's still very much WIP, not running yet, but figured I might as well have it open for visibility.


components/dada-execute/src/step.rs
```diff
@@ -0,0 +1,327 @@
...
+    ) -> Self {
+        Self {
+            db,
+            machine,
+            kernel,
+            temporaries: vec![],
+        }
+    }
+
+    pub(crate) fn step(&mut self) -> eyre::Result<ControlFlow> {
```
nikomatsakis, line 61: @brson 

This is the most fine-grained interface now: basically, take one step at a time. We should discuss how this fits with the REPL.

components/dada-execute/src/step/apply_op.rs
```diff
@@ -0,0 +1,106 @@
...
+            let span = self.span_from_bir(expr);
+            Err(error!(span, "divide by zero").eyre(self.db))
+        };
+        let overflow_error = || {
+            let span = self.span_from_bir(expr);
+            Err(error!(span, "overflow").eyre(self.db))
+        };
+        match (&self.machine[lhs], &self.machine[rhs]) {
+            (&ObjectData::Bool(lhs), &ObjectData::Bool(rhs)) => match op {
+                Op::EqualEqual => Ok(self.machine.our_value(lhs == rhs)),
```
brson, line 43: Why do these new values become "our" instead of "my"?

brson:
Still digesting this but it reads great. Thanks for writing excellent comments.

components/dada-execute/src/step/apply_op.rs
nikomatsakis: In general, scalar values should be "our" so that they get copied from place to place when they are used. "my" would be for things (like objects) that have identity.

(side note, I've been rethinking some of these names, but that's neither here nor there.)

nikomatsakis:
OK, the new interpreter and the playground works. Moving this out of draft, even though there are still some rough edges.

(One particularly confusing thing I noted is that the way I am printing thunks means that if you forget a "await" the output is ... not obvious.)

nikomatsakis:
Uh, so I accidentally pushed this again instead of instructing bors, but that's ok. =) 

(I have reset my git url to https to try and avoid this in the future...)

PR #106, opened 2022-02-04
track/check expected output from text

nikomatsakis:
This change extends the interpreter to track where each piece of output was created, and then extends the test harness to have annotations for expected output. This lets us better document the expected control flow from a test.

Another expected outcome (not yet implemented) is to highlight the text which was output from a breakpoint and to allow people to click in the output window to jump to the point in the program where the output was generated.

(@brson you've been poking at this code most recently)


nikomatsakis:
In tests, you can do:

```
#! OUTPUT ANY
```

to indicate you don't care to annotate, the output, or you can do

```
#! OUTPUT foo
```

to give a regular expression `foo` for the output that should be emitted at a particular line.

PR #115, opened 2022-02-08
Float support

### LINKED ISSUE #64: Interpret float binops
https://github.com/dada-lang/dada/pull/56 implements execution of int binops. Floats still need to be done. Requires consideration of how to handle float eq.

brson:
I found myself wanting floats for dogfooding the repl.

This is using the almost unused and unmaintained [eq_float](https://docs.rs/crate/eq-float/latest/source/src/lib.rs) crate to store Eq-implementing floats in the IR. I'm happy to do something else.

Fixes #64


brson:
This is missing a negative test case for "1." I'll add it later.

brson:
I added a test that parsing floats without decimals is an error. The span pointing to the dot is not quite right, but I don't know how to fix it offhand.

dada_tests/interpret/ops/ops_float.dada
```diff
@@ -0,0 +1,79 @@
...
+    var negneginf = 0.0 - neginf1
+    if inf1 == negneginf {
+        print("inf1 == negneginf").await
+        #! OUTPUT .*
+    }
+
+    # nan != nan
+    var nan1 = 0.0 / 0.0
+    var nan2 = 0.0 / 0.0
+    if nan1 == nan2 {
```
nikomatsakis, line 50: Man, I am so tempted to change this behavior and adopt IEEE total order for floats. =)

dada_tests/parser/float_no_decimal.dada
```diff
@@ -0,0 +1,5 @@
+fn main() {
+    a = 1.
```
nikomatsakis, line 2: Should we permit this? People sometimes do.

nikomatsakis (changes requested):
This seems good!

Can we add a test for `1_000.222_444` or something like that?

brson:
> This seems good!
> 
> Can we add a test for `1_000.222_444` or something like that?

I've added this test. Note the syntax supports trailing underscores like `1_.2_`. This is consistent with Rust.

components/dada-ir/src/code/syntax.rs
```diff
@@ -101,6 +101,9 @@ pub enum ExprData {
     /// `22`, `22_222`, etc
     IntegerLiteral(Word),
 
+    /// `2.2`
+    FloatLiteral(Word, Word),
```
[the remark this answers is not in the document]
brson: I've changed the comment to be in the spirit of your other comments on these enums, but it's a bit awkward.

dada_tests/interpret/ops/ops_float.dada
brson: I'll open an issue.

dada_tests/parser/float_no_decimal.dada
brson: I have no opinion.

nikomatsakis (approved):
Nice!

Issue #78, opened 2022-01-18
Add a repl


brson:
If dada is to be capable of interactive use then it needs to be architected and tested early. And some people (like me!) can usefully dogfood even a basic repl.

I think we should add a repl and get it under test in a way that it can maintain parity with the batch interpreter, probably put it on the playground.

I have never implemented a repl before, but to my mind the main thing that distinguishes a repl is that it maintains a persistent environment of local variables that can be referenced by future input lines, something like an endless stack frame. This is something that has always been hard for Rust, with a lot of efforts at repls simply growing a large buffer of statements that are continually reexecuted. It'll be the hard part of a dada repl too, but that's why we need to figure it out now.

To get started I think we can quickly achieve a few things:

- All the basic repl UI features of line reading, signal handling, history management, via rustyline
- Multiline parsing by scanning for open token trees
- Distinguishing between items and expressions
- Compilation of items by:
    - maintaining a map of item names to source code
    - each evaluation, recreating a single source text containing all the items and updating the db.
 - Compilation of expressions by:
     - Wrapping them in a function and interpreting that function

It's going to take a lot of experimentation to figure out how to deal with expressions.

I think the repl will need to distinguish between variable assignment expressions and non-variable-assignment expressions, but even non-variable-assignment expressions will have to reference previous bindings.

@nikomatsakis has suggested a pattern of code generation that should work well to get started with expressions that reference variables, while leaning hard on dada's incremental compilation:

Given this sequence of inputs:

```
> x = 22
> y = x + 22
```

Transform it into:

```
async fn step1() {
    x = 22
    step2(x).await
}

async fn step2(x) {
    y = x + 22
    step3(x, y).await
}

async fn step3(x, y) {
    y
}
```

For every step the generated function simply passes on every variable to a stub function for the next step, preserving all the variable bindings for the next step's use. When the next step evaluates it replaces its stub function with its own definition and creates a stub for the next-next step.

dada's incremental compilation should make this quite efficient without much effort.

There are plenty of problems to figure out with this pattern, but it's worth starting and seeing how far it can go.

For testing, we'll probably add a mode that actually starts another process and pipes stdin/out. We should be able to make it so all batch tests that compile and run successfully can be fed into the repl and behave the same, using the same reference files, though confirming that both modes issue the same e.g. warning diagnostics could be tricky. It may also be easy use the same reference files for files that compile successfully but have runtime errors. Compiler errors could be harder, not sure.

brson:
[...]

PR #125, opened 2022-02-11
[wip] repl implementation

brson:
This is getting to the point where the features I have been aiming for basically work and the code is clean enough to read. But it's not really reliable enough to use yet.

There is a new crate called dada-repl that implements the repl behavior, minus the I/O and kernel. It is divided into two modules: `read`, that figures out what a single line represents, and manages the buffering of multiline inputs; and `eval` that manages a growing source module and executes expressions by turning them into specially-crafted functions.

The basic strategy is as we have discussed in https://github.com/dada-lang/dada/issues/78. It builds up an ever-growing source file in a way that is intended to minimize recompilation; and during evaluation it pauses the interpreter after the end of a step, prints the result, then replaces the top stack frame to continue on the next step.

Source produced by the repl looks like:

```
# Items come first

# >>> fn five() { 5 }
fn five() { 5 }

# Wrapped expressions come next

# >>> five()
# 5
__repl_expr_fn_0(__repl_result) {
    __repl_result = five()
    __repl_expr_fn_1(__repl_result)
}

# >>> a = 6
__repl_expr_fn_1(__repl_result) {
    a = 6
    __repl_result = ()
    __repl_expr_fn_2(__repl_result, a)
}

# >>> a + foo()
# 11
__repl_expr_fn_2(__repl_result, a) {
    __repl_result = a + foo()
    __repl_expr_fn_3(__repl_result, a)
}

# stub for next step
__repl_expr_fn_3(__repl_result, a) {
    # execution paused here and this stack frame will be replaced
}
```

There is a CLI driver in dada-lang/src/repl.rs. It is intended that there could be a wasm driver as well.

It is intended that the dada-repl crate doesn't do any I/O on its own but now it is printing diagnostics, as well as the result of expressions.

It is intended that the repl has minimal impact on other crates but as of now I've hacked some things into other crates as needed to get access for the repl.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -141,6 +141,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "clipboard-win"
+version = "4.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2f3e1238132dc01f081e1cbb9dace14e5ef4c3a51ee244bd982275fb514605db"
+dependencies = [
+ "error-code",
+ "str-buf",
+ "winapi",
+]
+
+[[package]]
 name = "console_error_panic_hook"
 version = "0.1.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -359,12 +370,14 @@ dependencies = [
  "dada-lex",
  "dada-lsp",
  "dada-parse",
+ "dada-repl",
  "eyre",
  "ignore",
  "lsp-server",
  "lsp-types",
  "parking_lot",
  "regex",
+ "rustyline",
  "salsa",
  "serde",
  "serde_json",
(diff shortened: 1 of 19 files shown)


nikomatsakis:
I was wondering whether it is worth adding a new top-level kind of `Item`, or perhaps a sub-species of `Function`, something like that. I suspect probably yes. 

Similarly, I've given some more thought to it, and I think I would like to move the repl storage into salsa -- it occurs to me that the interpreter could have a query that wraps it. The reason for this is that interpreter values contain e.g. references to entities like `Class`, and strictly speaking those should not be persisted across mutations except as part of stored query values. 

I think these changes would likely be "skin deep" to the work you've done in any case. I have to read it more carefully to make specific suggestions.

brson:
I've made a few improvements recently:

- Rollback any steps that produce an error
- Track and ignore previously-seen diagnostics

I've also started using this for basic calculator features.

Regarding chunk-of-code items, I think it would be the biggest help to a repl if it can incrementally recomputed. It might make sense to have code outside of functions, but I would expect some way to then distinguish between executable modules and library modules - having arbitrary code run on module load ala python is too unpredictable for my taste.

brson:
I've also added a `:load` repl command, a simple hack to load dada files into the repl.

nikomatsakis:
This is cool :)

brson:
This is bitrotted and I am not going to get back to it soon.

PR #129, opened 2022-02-11
Support signed int

### LINKED ISSUE #62: Add integer suffixes to the parser
Right now there is compiler support for signed int types, but no way to create them. Per https://github.com/dada-lang/dada/pull/56#issuecomment-1013675589, add Rust-style integer suffixes to the parser so it can produce signed int types.

@nikomatsakis what suffixes do you want for these? "uint" / "int", "u64", "i64"?

XFFXFF:
close #62 
Add integer suffixes to the parser, `i` for signed int, and `u` for unsigned int.


components/dada-ir/src/code/syntax.rs
```diff
@@ -98,8 +98,8 @@ pub enum ExprData {
     /// true, false
     BooleanLiteral(bool),
 
-    /// `22`, `22_222`, etc
-    IntegerLiteral(Word),
+    /// (`22i`, suffix_start: 2), (`22_222u`, suffix_start: 6), etc
+    IntegerLiteral(Word, usize),
```
nikomatsakis, line 102: ...then this could be `IntegerLiteral(Word, Word)`, too, which seems mildly nicer.

components/dada-lex/src/lex.rs
```diff
@@ -130,8 +130,20 @@ where
                     push_token(Token::Comma);
                 }
                 '0'..='9' => {
-                    let text = self.accumulate(ch, |c| matches!(c, '0'..='9' | '_'));
-                    push_token(Token::Number(text));
+                    let mut text = self.accumulate_string(ch, |c| matches!(c, '0'..='9' | '_'));
```
nikomatsakis, line 133: ...no changes here required

components/dada-parse/src/parser/code.rs
```diff
@@ -312,13 +312,13 @@ impl CodeParser<'_, '_> {
...
-        } else if let Some((word_span, word)) = self.eat(Number) {
+        } else if let Some((word_span, (word, suffix_start))) = self.eat(Number) {
             let whitespace_before_dot = self.tokens.skipped_any();
             match self.eat_op(Op::Dot) {
-                None => Some(self.add(ExprData::IntegerLiteral(word), word_span)),
+                None => Some(self.add(ExprData::IntegerLiteral(word, suffix_start), word_span)),
                 Some(dot_span) => {
                     let whitespace_after_dot = self.tokens.skipped_any();
-                    if let Some((_, dec_word)) = self.eat(Number) {
+                    if let Some((_, (dec_word, _))) = self.eat(Number) {
```
nikomatsakis, line 321: No changes *here* required (i.e., `22i.34` isn't something we could parse)

components/dada-validate/src/validate/validator.rs
```diff
@@ -172,17 +172,46 @@ impl<'me> Validator<'me> {
...
+                                self.span(expr),
+                                "`{}` is not a valid integer: {}",
+                                w.as_str(self.db),
+                                e,
+                            )
+                            .emit(self.db);
+                            self.add(validated::ExprData::Error, expr)
+                        }
+                    },
+                    "" | "i" => match i64::from_str(&without_underscore) {
```
nikomatsakis, line 197: Hmm. I'm almost tempted to say that unsigned should be the default! But I guess that's nonsense.

components/dada-execute/src/step.rs
```diff
@@ -386,6 +386,10 @@ impl<'me> Stepper<'me> {
                 permission: self.machine.new_permission(ValidPermissionData::our()),
             }),
             bir::ExprData::IntegerLiteral(v) => Ok(Value {
+                object: self.machine.new_object(ObjectData::Int(*v)),
+                permission: self.machine.new_permission(ValidPermissionData::our()),
+            }),
+            bir::ExprData::UnsignedIntegerLiteral(v) => Ok(Value {
```
nikomatsakis, line 392: We have to modify `Stepper::apply_op` to determine what to do when (e.g.) two uints are added, or a uint plus an int, etc.

I am debating what the rules should be here!

I am thinking *maybe* we want to distinguish:

* signed integer (written `0i`)
* unsigned integer (written `0u`)
* unsuffixed integer (written `0`)

The idea would be that signed/unsigned can only mix with each other, but unsuffixed integers can mix with either.

We would then have some way to convert from signed to unsigned and vice versa.

My idea is this: I'm trying to move the rules that a static type system would impose to runtime.

components/dada-parse/src/parser/code.rs
```diff
@@ -312,13 +312,13 @@ impl CodeParser<'_, '_> {
         } else if let Some((id_span, id)) = self.eat(Identifier) {
```
nikomatsakis, line 312: We are going to want to support `-` and `+` as unary operators, I imagine, so that we can parse this:

```
x = -5
y = -(1)
z = -x
```

We might want to do something mildly unconventional and parse `-5` without any whitespace as a literal that is explicitly a signed integer.

components/dada-parse/src/parser/code.rs
```diff
@@ -312,13 +312,13 @@ impl CodeParser<'_, '_> {
...
             tracing::debug!("identifier");
             Some(self.add(ExprData::Id(id), id_span))
```
nikomatsakis, line 314: We want to be careful and look at the binary operator parsing; we can push that to a follow-up PR, but let's at least add tests here (with `#! FIXME`). The question is how to think about binary operators and newlines. Rust had the same issue to wrestle with and I suspect we want the same general answer. I'm referring to things like this:

```
fn foo() -> {
    if true { 1 } else { 2 }
    -5 # probably wants to return `-5`, not `-4`
}

fn foo() -> {
    a = if true { 1 } else { 2 }
    -5 # probably wants to return `-5` and set `a` to 1
}

fn foo() -> {
    a = (if true { 1 } else { 2 }
    -5) # probably wants to set `a` to `-4` and return `()`? Not sure.
}

fn foo() -> {
    a = if false { 1 } else { 2
    -5} # probably wants to set `a` to `-3`, but I'm not entirely sure ,especially since the next example...
}

async fn foo() -> {
    a = if false { 1 } else { print(2).await
    -5} # ...probably wants to print the number `2` and then set `a` to `-5`, and not try to subtract `5` from `()`
}
```

I'm somewhat tempted to just say "binary operators cannot be preceded by a newline", so that you have to write `b - \n 5` and not `b \n - 5`. That'd be a very simple rule.

nikomatsakis: But opening a Question: issue to hash it out seems good.

nikomatsakis (changes requested):

XFFXFF:
@nikomatsakis  Thanks for the patience guidance, I will change it after I digest these. :)

components/dada-parse/src/parser/code.rs
nikomatsakis: Well, we could code it up either way. I'm honestly not sure if it *should* be parsed as a binary operator! It sure *looks* like a missing comma to me. =)

components/dada-execute/src/machine.rs
```diff
@@ -198,6 +200,7 @@ impl ObjectData {
             ObjectData::Tuple(_) => "a tuple".to_string(),
             ObjectData::Bool(_) => "a boolean".to_string(),
             ObjectData::Uint(_) => "an unsigned integer".to_string(),
+            ObjectData::UnsuffixedInt(_) => "an unsuffixed integer".to_string(),
             ObjectData::Int(_) => "an integer".to_string(),
```
nikomatsakis, line 204:
```suggestion
            ObjectData::Int(_) => "a signed integer".to_string(),
```

components/dada-execute/src/step/apply_op.rs
```diff
@@ -43,7 +43,10 @@ impl Stepper<'_> {
                 Op::EqualEqual => Ok(self.machine.our_value(lhs == rhs)),
                 _ => op_error(),
             },
-            (&ObjectData::Uint(lhs), &ObjectData::Uint(rhs)) => match op {
+            (&ObjectData::Uint(lhs), &ObjectData::Uint(rhs))
+            | (&ObjectData::Uint(lhs), &ObjectData::UnsuffixedInt(rhs))
+            | (&ObjectData::UnsuffixedInt(lhs), &ObjectData::Uint(rhs))
+            | (&ObjectData::UnsuffixedInt(lhs), &ObjectData::UnsuffixedInt(rhs)) => match op {
```
nikomatsakis, line 49: debatable how two unsuffixed integers should behave... I think I'd like `5 - 3` to result in an unsuffixed integer and `3 - 5` to fail at runtime (underflow). What do we do now?

nikomatsakis (approved):
Gave this a quick skim, looks like it's going in a good direction. Left a comment.

components/dada-execute/src/step/apply_op.rs
XFFXFF: Now `3 - 5` fails at runtime (underflow), but `5 - 3` results in an unsigned integer. A `value` of type u64 is implicitly converted to ObjectData::Uint when we call `self.machine.our_value(value)` 
https://github.com/dada-lang/dada/blob/547f5166f525cfa4bb61e9fa0cde29dbc5a406b1/components/dada-execute/src/machine.rs#L209-L235

components/dada-execute/src/step/apply_op.rs
```diff
@@ -67,31 +69,35 @@ impl Stepper<'_> {
             (&ObjectData::Int(lhs), &ObjectData::Int(rhs)) => match op {
                 Op::EqualEqual => Ok(self.machine.our_value(lhs == rhs)),
                 Op::Plus => match lhs.checked_add(rhs) {
-                    Some(value) => Ok(self.machine.our_value(value)),
+                    Some(value) => Ok(self.machine.our_value(ObjectData::Int(value))),
                     None => overflow_error(),
```
XFFXFF, line 73: I renamed `Int` to `SignedInt` and `UnsuffixedInt` to `Int`. I also changed the way two unsuffixed integers behave, now `5 - 3` results in an unsuffixed integer.

dada_tests/interpret/ops/op_neg_uint.dada
```diff
@@ -0,0 +1,5 @@
+fn main() {
+    -1u
+#! ^^^^ RUN ERROR cannot apply operator - to an unsigned integer
+# FIXME: Error highlighting should not include the leading whitespace
```
XFFXFF, line 4: The span of the unary expression includes the leading whitespace, which is weird

components/dada-parse/src/parser/code.rs
```diff
@@ -325,15 +332,27 @@ impl CodeParser<'_, '_> {
...
         } else if let Some((word_span, word)) = self.eat(Number) {
```
XFFXFF, line 334: > We might want to do something mildly unconventional and parse -5 without any whitespace as a literal that is explicitly a signed integer.  

I don't know how to implement this. I can't do this in `parse_expr_0`, because `-` will always be consumed in either unary or binary expressions.

dada_tests/interpret/ops/op_neg_uint.dada
XFFXFF: I think I have fixed it

components/dada-parse/src/parser.rs
```diff
@@ -63,7 +63,7 @@ impl<'me> Parser<'me> {
     /// if so, return the span and the tokens after skipping the operator.
     fn test_op(&self, op: Op) -> Option<(Span, Tokens<'me>)> {
         let mut tokens = self.tokens;
-        let span0 = tokens.last_span();
+        let span0 = tokens.peek_span();
```
XFFXFF, line 66: This to fix the span bug I mentioned in https://github.com/dada-lang/dada/pull/129#discussion_r806819338.  
```
fn main() {
   - 1
}
```
When calling `self.test_op(Op::Minus)`, `tokens.last_span` returns the span of `{`, and `tokens.peek_span()` returns the span of `-`

components/dada-execute/src/machine/stringify.rs
```diff
@@ -19,9 +19,10 @@ pub(crate) impl<T: ?Sized + MachineOp> DefaultStringify for T {
         match &self[value.object] {
             ObjectData::String(s) => s.to_string(),
             ObjectData::Bool(v) => format!("{}", v),
-            ObjectData::Int(v) => format!("{}", v),
+            ObjectData::SignedInt(v) => format!("{}", v),
```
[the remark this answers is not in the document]
nikomatsakis: or just `{}i`, depending on what we want to recommend I guess

nikomatsakis:
Another test:

```
a = (22 + 44) # still has no suffix
print(a + 22i) #! OUTPUT 88i
print(a + 22u) #! OUTPUT 88u
```

(I can't remember if that's what would happen...? I think so)

nikomatsakis:
Probably good to have a test or two that uses `*`, `/`, and other operators beyond `+` to combine unsuffixed and suffixed integers, too.

Also, test that we can parse:

```
x = a - -b
```

```
x = a - - b
```

```
x = a -      # possibly this should be an error :)
-b
```

and let's do a test a precedence related test

```
print(22 + -66 / 3).await #! OUTPUT 44
```

finally some tests that combine negative unsuffixed with a `u`

```
a = -66 + 3u #! RUN ERROR foo
```

I'm not entirely sure what this next one should do =) probably an error?

```
a = -22 + 44u #! RUN ERROR foo
```

nikomatsakis:
Hey @XFFXFF, left a bunch of tests and things, but overall this seems to be really close!

nikomatsakis (approved):

PR #130, opened 2022-02-11
support for pasing `return`

### LINKED ISSUE #128: Return is not parsed
Just noticed that `return` is not parsed. We should fix that! I think the validation/BIR code is mostly there, but I guess we'll find out.

### Expected syntax

```
return <expr>
```

### Precedence

The precedence should be very low, so that 

```
return 22 + 44
```

returns 66 etc

XFFXFF:
fixes #128 
Right now an expr is required after `return`, should we support return empty?
```
fn main() {
  return
}
```


components/dada-ir/src/code/syntax.rs
```diff
@@ -161,6 +161,9 @@ pub enum ExprData {
     /// `a := b`
     Assign(Expr, Expr),
 
+    /// return
+    Return(Expr),
```
nikomatsakis, line 165: We do want to support `return` by itself. The syntax tree would have `Option<Expr>`, I imagine, but the validator would substitute a `()` expression.

components/dada-parse/src/parser/code.rs
```diff
@@ -312,6 +312,15 @@ impl CodeParser<'_, '_> {
         } else if let Some((id_span, id)) = self.eat(Identifier) {
             tracing::debug!("identifier");
             Some(self.add(ExprData::Id(id), id_span))
+        } else if let Some((return_span, _)) = self.eat(Keyword::Return) {
```
nikomatsakis, line 315: I'm debating if `parse_expr_0` is the right place for this. I think I would put it at the highest level (e.g., `parse_expr` itself).

As an example of why, consider this (and we should make this into a test)

```
fn foo() {
    return * 3
}
```

I think I want that to be a parse error, but I believe it will parse with the setup you have here. The same is true for any binary operator that cannot begin an expression (e.g., `return && true`).

nikomatsakis (changes requested):
Nice :) Some comments.

PR #136, opened 2022-02-19
Fix typos in the book

matheus-consoli:
Just fixes a bunch of typos found in the book.

I found just the first and decide to run [typos-cli](https://github.com/crate-ci/typos), this PR is not a big deal, and I'm yet reading the book. Feel free to not merge this if you will.

`typos` also complained about some documentation and variable typos, let me know if you want me to fix it as well.

PR #137, opened 2022-02-21
Fix code typos

matheus-consoli:
The spiritual successor of #136

### DIFF
--- a/components/dada-breakpoint/src/breakpoint.rs
+++ b/components/dada-breakpoint/src/breakpoint.rs
@@ -59,7 +59,7 @@ pub fn find_item(db: &dyn crate::Db, filename: Filename, offset: Offset) -> Opti
 /// that is, the moment when all of E's children have been
 /// evaluated, but E has not yet taken effect itself.
 ///
-/// Assumes: the offest is somewhere in this syntax tree.
+/// Assumes: the offset is somewhere in this syntax tree.
 ///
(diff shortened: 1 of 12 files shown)


components/dada-execute/src/heap_graph.rs
```diff
@@ -51,7 +51,7 @@ tables! {
     pub(crate) struct Tables {
         stack_frame_nodes: alloc StackFrameNode => StackFrameNodeData,
         objects: alloc ObjectNode => ObjectNodeData,
-        datas: alloc DataNode => DataNodeData,
```
XFFXFF, line 54: Not sure if this is a typo

matheus-consoli: yeah, I also was in doubt about this one... data is already in plural form, from my understanding, and `datas` don't really is a word. Even so, I have done and undone this specific some times.
If it's a design choice, let me know and I change it back.

brson: Yeah, I think this is an intentional misspelling for naming consistency. I think it's best to leave it.

brson:
Thanks for teh spell checking!

components/dada-execute/src/heap_graph.rs
matheus-consoli: Makes sense to me! Removed it!

XFFXFF:
Thanks!  
bors r+

Issue #139, opened 2022-02-27
Modify test harness to stop emitting color; add IR reference files to gitignore


nikomatsakis:
Per https://github.com/dada-lang/dada/issues/135, we should modify the test harness...

* to stop emitting color in the stdout files
* to generate the various IR files but not to check if they have changed (they are for looking at, not for reference)

The second step allows us to add them to `gitignore` (and remove the existing files from the repository)

nikomatsakis:
Actually, @lqd, in merging #140 I overlooked an issue!

That PR ignored *all* ref files, but I think we still want to be checking certain files-- actually, I guess it's just one.

Want to check:

* annotations in file match actual errors (:heavy_check_mark: this matches main)
* stdout/stderr from execution match `stdout.ref` and `stderr.ref` (and matches annotation in file) (:heavy_check_mark: this matches main, I believe)
* compiler output matches `ref.ref` file (:x: we don't check this on main now -- also, we should rename `ref.ref`)

Do not want to check (or commit into the repo), but do want to generate:

* IR generated (syntax.ref, validated.ref, bir.ref)
* LSP output (`lsp.ref`)

Maybe we should rename those files to `.debug` or something?

lqd:
I'll take care of that

nikomatsakis:
Oh, I forgot one thing -- `HeapGraph-*.ref` -- I'm torn on this one. I do want to be testing that the heap graph dump is successful and contains certain key elements, but the ref files contain an awful lot of details.

Still, I think for now we should keep them.

nikomatsakis:
The idea is that they are testing that the `#?` output contains what we expect it to contain. Admittedly, I mostly ignore diffs that occur in them, which is a bad sign. :)

nikomatsakis:
(One thing I was considering was trying to find some kind of "ascii-art" renderer for graphviz files and checking *that* in, I think I would read that much more happily...)

nikomatsakis:
@lqd yes! The only tools I found for that were written in perl, which was kind of a drag.

nikomatsakis:
Honestly, maybe we should do a web request to that site :)

nikomatsakis:
(heh, or I guess setup our own...that'd be kind of rude)

nikomatsakis:
I wonder if we could compile the whole kit-and-kaboodle to wasm somehow :)

nikomatsakis:
@lqd filed https://github.com/dada-lang/dada/issues/150

[...]

PR #140, opened 2022-02-27
Update test harness and handling .reference files

lqd:
This PR fixes (hopefully) #139:
- by introducing control over colors when formatting errors
- making the test harness not use colors when formatting diagnostics: the very readable result is visible in the "bless colorless test expectations" commit.
- if I understood #139 correctly: all .ref files were to be "ignored" during tests (except when blessing, where they are still generated), removed from the repo, and `gitignore`d ? I hope this understanding is correct, as the last few commits do that.


components/dada-lang/src/test_harness.rs
```diff
@@ -166,35 +166,31 @@ impl Options {
             &expected_diagnostics.compile,
             &mut errors,
         )?;
-        self.check_output_against_ref_file(
+        self.maybe_bless_ref_file(
```
nikomatsakis, line 169: I was going to suggest that we make this *always* bless, but maybe it's good to have a mode that doesn't generate output files. (I somewhat wonder if `--bless` should be the default, even, and instead we have `--check`, but I guess that could be surprising.)

nikomatsakis (approved):

nikomatsakis:
Thanks @lqd!

bors r+

components/dada-lang/src/test_harness.rs
lqd: Could be a good idea to always bless by default and ensure you're not looking at old files.

nikomatsakis: If you wanted to open up a follow-up PR, would be welcome. Maybe a good idea would be to *try* and update the ref files, but if an I/O error occurs, just *warn* about it and don't fail the tes.t

PR #143, opened 2022-02-28
modal harmony tutorial

nikomatsakis:
This is a "WIP" that tries to explain the [new "modal harmony" branch](https://github.com/nikomatsakis/dada/tree/modal-harmony) I've been working on. I'm opening it to let folks see the preview and/or leave comments/thoughts.


book/src/dyn_tutorial/permissions.md
```diff
@@ -0,0 +1,126 @@
...
+If you don't specify any ownership permision, the default is `our leased`. This means that the variable cannot be used to mutate the fields of the object. Example:
+
+```
+class Point(x, y)
+
+async fn main() {
+    p = Point(22, 44)      # declares `p` in `our leased` mode
+    p.x += 1               # Error!
+}
+```
```
XFFXFF, line 91: I wonder why the default permission is `our leased`, not `our`.  
From my previous understanding of dada's ownership, the owner (unique or shared) has the responsibility for freeing the object, but the variable with `our leased` permission is not the owner of the object. So who is responsible for freeing the `Point(22, 44)`, is it `p`? 
```
class Point(x, y)
async fn main() {
    p = Point(22, 44)
}
```
Just thought it would make sense if this code was interpreted as 
```
class Point(x, y)
async fn main() {
    my tmp = Point(22, 44)
    p = tmp.lease.share
}
```
`Point(22, 44)` are freed once `tmp` has gone out of scope. I'm not sure if this is the correct way to understand.

nikomatsakis: Good question! In Rust, terms, `our leased` is effectively `&T`, whereas `our` is effectively `Rc<T>`. I think the more common scenario is that functions take `&T` arguments (which they will use but not "capture") than `Rc<T>`.

To use a more dada-like answer, it is also affected by the subtyping and other rules (which I haven't described in the tutorial yet). The idea is that when you assign to a variable, if the variable demands owned content (`my`, `our`), you must give it something owned; if it demands unique access (`my`, `leased`), you must give it something unique. But if it demands leased, you can give it owned *or* leased, and similarly, you can assign a uniquely accessible thing into a shared location (it becomes shared). You can think of this as a kind of coercion:

* `my` coercible to `our`
* `my` coercible to `leased`
* `our` coercible to `our leased`
* `leased` coercible to `our leased`

By using `our leased` as the default, it means then that you can assign to that variable with virtually any value. This makes sense since "our leased" is basically saying "give me something I can read; I promise not to hold on to it", and that could be any of the permissions.

book/src/dyn_tutorial/permissions.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Ownership permissions are enforced
+
+We saw that `my p` indicates that the variable `p` has unique ownership of its contents. But what happens if we try to store a value in `p` that cannot be owned? As an example, integers like `22` and `44` are inherently shared. If we try to claim ownership of those, we get an error. As an example, try running this:
```
brson, line 95: The idea that integers are inherently shared is not obvious to me. You've explained it me previously but I can't recollect the reasoning. It might be good to expand on that.

Intuition tells me I should be able to own and mutate any value.

brson: One additional consideration here for me - I know the sharing mechanism hasn't been explained in the text, but considering that `our` types are meant to be reference counted, this suggests to be that integers are reference counted. I doubt that is the case, but I'm curious what distinguishes integers from other types to avoid putting them on the heap.

book/src/dyn_tutorial/permissions.md
```diff
@@ -0,0 +1,126 @@
...
+}
+```
+
+Now try changing the program to use `our x = 22`: you'll find that it works. You can also write `x = 22`, which is short for `our leased x = 22`. All of these work fine. This is because integers like `22` are shared: you cannot assert unique access to them. 
+
+**Exercise:** What do you think will happen with `leased x = 22` or `any x = 22`? Give it a try and see!
+
+## Ownership permissions on fields
+
+Ownership permissions also apply to fields and function arguments. The declaration of `Point` we have been using actually declares its fields with the default permissions, `our leased`:
```
brson, line 109: This would imply that structs with default-permissioned fields are tied to a stack discipline, and one would need to write all `my` and `our` fields to create a type that can be put on the heap without being tied to stack lifetimes?

book/src/dyn_tutorial/permissions.md
```diff
@@ -0,0 +1,126 @@
...
+# class Point(our leased x, our leased y)
+```
+
+Just as with local variables, declaring a field to be `my` means that the class can only store values that it has unique ownership of:
+
+```
+class Point(my x, my y)
+
+async fn main() {
+    my p = Point(22, 44)      # Error!
```
brson, line 124: What's the syntax to take ownership of the integers here and fix the error?

book/src/dyn_tutorial/permissions.md
```diff
@@ -0,0 +1,126 @@
...
+}
+```
+
+## Ownership permissions are enforced
+
+We saw that `my p` indicates that the variable `p` has unique ownership of its contents. But what happens if we try to store a value in `p` that cannot be owned? As an example, integers like `22` and `44` are inherently shared. If we try to claim ownership of those, we get an error. As an example, try running this:
+
+```
+async fn main() {
+    my x = 22               # Error!
```
brson, line 99: What's the syntax to take ownership of a `22` here?

brson: Maybe it's just not possible? I can create an `our` and reassign the `our` as a way to "mutate" an integer, but I can't ever have a `my Int`?

book/src/dyn_tutorial/permissions.md
nikomatsakis: `our x`

book/src/dyn_tutorial/permissions.md
nikomatsakis: Yes. You can do `our x = 22`, which then permits you to do (e.g.) `x += 1`, but the "fields of the integer" are not changed. Think of like Java's `Integer` class.

book/src/dyn_tutorial/permissions.md
nikomatsakis: I don't intend to ref-count every integer, no, but I *do* intend that you can *think of things* as having every integer be a ref-counted object (much like it would be in Python).

book/src/dyn_tutorial/permissions.md
nikomatsakis: Yes, basically. My assumption is that `my` and `our` will be used all over the place on fields.

I'm not sure if this will feel good or not! I feel quite good about `my` and `our` on variables, and they are quite short, so I wanted to go for total consistency across all kinds of variables.

One other thought I've had is that the default for classes might be 'parametric', i.e., the owner of the class can 'raise up" the minimum level of ownership. But that's kind of the opposite of how modes typically work, where they "lower down" the full permissions (e.g., if you have `class Pair(my a, my b)` and then you have a `our pair: Pair`, you only have shared access to `pair.a` and `pair.b`).

book/src/dyn_tutorial/permissions.md
nikomatsakis: > Intuition tells me I should be able to own and mutate any value.

Just to be clear, you can own the value, you just can't mutate an integer's fields (it doesn't have any). It's like you own an `Rc<i32>`, essentially. 

Maybe it helps for me to say that I expected to have `our class Point(our x, our y)` as a way to declare your own class that is "always shared". In this case, you could never have a `my p = Point(22, 44)`, because nobody can have single ownership of a `Point`; you could have `our p = Point(22, 44)`, and you could also do `p := Point(44, 66)` to change that to another point.

brson: Thanks. I think I understand. I need to stop trying to map Rust's `let` and `let mut` onto Dada's permissions.

book/src/dyn_tutorial/permissions.md
brson: Starting with consistency is best.

book/src/dyn_tutorial/permissions.md
brson: This suggests to me there are no immutable bindings in dada? Is that correct?

Thanks for all the clarifications.

book/src/dyn_tutorial/permissions.md
pro465: @nikomatsakis 

> Good question! In Rust, terms, `our leased` is effectively `&T` [...]

shouldn't it be `&mut T`?

nice language btw, I'm excited about the future of dada!

nikomatsakis: > shouldn't it be `Rc<&mut T>`?

Interesting question! I wasn't thinking of `our leased` as `our (leased T)` but more like `(our leased) T` -- in other words, it's like a lease that we collectively own. That said, I think we might want to make the concept of a "shared lease" more first class for other reasons, and I was thinking of using the keyword `shlease` for it (shared-lease, pronounced "shlease"), in which case this could be `shleased x = ...` I suppose. I'm not sure if I can tolerate that keyword but it's kind of amusing. 

I don't like the discrepancy between `our` and `shlease` though, nor do I love the discrepancy between `our` and `share`. It might be better to call `our` just `shared`. Then we'd have

* my
* shared
* leased
* shleased

`my` sort of stands out here, but `owned` just seems too long. =) And I did like the pithiness and playfullness of `our`!

nikomatsakis:
I've been a bit busy with other things lately but I sat down and worked through the "outline" for the full tutorial as I imagine it. Should be visible in the preview soon:

https://deploy-preview-143--dada-lang.netlify.app/tutorial_structure/dynamic_dada.html

book/src/dyn_tutorial/permissions.md
pro465: in my opinion, `owned` sounds better as a rust enthusiast.

book/src/dyn_tutorial/any.md
```diff
@@ -0,0 +1,71 @@
...
+Rather than labeling variables as `my` or `our`, you can also use the `any` keyword. This will permit the variable to store an object with any permission. When using `any`, the `give` and `share` keywords allow you to control the ownership:
+
+```
+class Point(our x, our y)
+
+# The point is `my` when first created
+any my_p = Point(22, 44)
+
+# You can `give` it to another variable
+any no_my_p = owned_p.give
```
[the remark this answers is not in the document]
nikomatsakis: thanks! pushed some fixes. 

Also, the name `no_my_p` was meant to be funny, kind of like "No! My p!", but re-reading it, it didn't work for me, so I changed it to `my_p_now`, which maybe works better...

PR #152, opened 2022-03-07
Test Harness take 2

lqd:
This PR changes the test harness to fix issues in the previous PR, and match the state described in https://github.com/dada-lang/dada/issues/139#issuecomment-1057856720:

I've brought back the `ref.ref` files (renamed to `compiler-output.ref` as suggested), the `stdout.ref` files, and HeapGraph dumps (until we figure out a possible ASCII-art graph pipeline).

As a summary:
- `.ref` reference files are important things: they need to be checked against, are therefore committed to the repo and updated on demand via `--bless`
- some of the intermediate files are useful for debugging only: they are renamed `.debug`, git-ignored, and always generated when running tests. Their output won't be stale when running and debugging tests.


nikomatsakis (approved):

nikomatsakis:
Thanks @lqd, you rock :guitar:

lqd:
This clippy warning seems new and wrong. It's surely triggering on master already.

```
Error:   --> components/dada-parse/src/parser.rs:54:9
   |
54 | /         if let Some(r) = op(self) {
55 | |             Some(r)
56 | |         } else {
57 | |             self.tokens = tokens;
58 | |             None
59 | |         }
   | |_________^ help: replace it with: `op(self)`
   |
   = note: `-D clippy::needless-match` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_match
```

nikomatsakis:
@lqd that does seem like a bug in clippy, I wonder if it's already been filed

lqd:
@nikomatsakis I'm opening a PR to work around it as we speak

PR #153, opened 2022-03-13
Introduce a react-based web app for dada

ciyer:
This PR introduces a react-based scaffolding for a web dada IDE.

image

PR #154, opened 2022-03-15
Work around an eager clippy warning

lqd:
As seen in https://github.com/dada-lang/dada/pull/152#issuecomment-1067710475 and https://github.com/dada-lang/dada/pull/153#issuecomment-1067708309, it seems a recent nightly has introduced a false positive in the `needless_match` lint, causing CI to fail as it denies warnings.

Just for reference, it looks side-effects in branches are missed, and thought to be replaceable by the matched Option.

```rust
pub fn side_effect() -> Option<String> {
    if let Some(r) = op() {
        Some(r)
    } else {
        println!("side-effect");
        None
    }
}
```

We could ignore the lint, much like other lints are sometimes ignored, or work around it like this PR. Which do you prefer ?

### DIFF
--- a/components/dada-parse/src/parser.rs
+++ b/components/dada-parse/src/parser.rs
@@ -48,15 +48,15 @@ impl<'me> Parser<'me> {
 
     /// Run `op` -- if it returns `None`, then no tokens are consumed.
     /// If it returns `Some`, then the tokens are consumed.
-    /// use sparingly, and try not to report errors or have side-effects in `op`.
+    /// Use sparingly, and try not to report errors or have side-effects in `op`.
     fn lookahead<R>(&mut self, op: impl FnOnce(&mut Self) -> Option<R>) -> Option<R> {
         let tokens = self.tokens;
-        if let Some(r) = op(self) {
-            Some(r)
-        } else {
+        let r = op(self);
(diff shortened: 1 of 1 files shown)


nikomatsakis:
This looks good.

bors r+

lqd:
That's quite a streak we got there, now an ICE from #95015 :)

```
thread 'rustc' panicked at 'assertion failed: !value.has_escaping_bound_vars()', /rustc/52b34550aca5f7dd7e152f773e3ab786acb86f6f/compiler/rustc_middle/src/ty/sty.rs:1089:9
query stack during panic:
#0 [is_sized_raw] computing whether `dyn std::io::Write` is `Sized`
#1 [layout_of] computing layout of `*const dyn std::io::Write`
#2 [layout_of] computing layout of `core::ptr::unique::Unique<dyn std::io::Write>`
#3 [layout_of] computing layout of `alloc::boxed::Box<dyn std::io::Write>`
```

lqd:
This will be fixed in tomorrow's nightly

nikomatsakis:
ok :)

nikomatsakis:
bors r+

let's try again in case we're lucky

Issue #142, opened 2022-02-28
`>=` and `<=` are not supported


XFFXFF:
`+=`, `-=`, `*=`, `/=` are supported in #83, but `>=` and `<=` are not.

Veeupup:
@XFFXFF  I'm verry interested in this issue, maybe let me try it? : D

PR #157, opened 2022-04-04
feature: support `>=` and `<=` operators

Veeupup:
Signed-off-by: Veeupup <code@tanweime.com>

fix issue: #142


components/dada-execute/src/step/apply_op.rs
```diff
@@ -41,12 +41,16 @@ impl Stepper<'_> {
         match (&self.machine[lhs], &self.machine[rhs]) {
             (&ObjectData::Bool(lhs), &ObjectData::Bool(rhs)) => match op {
                 Op::EqualEqual => Ok(self.machine.our_value(lhs == rhs)),
+                Op::GreaterEqual => Ok(self.machine.our_value(lhs >= rhs)),
+                Op::LessEqual => Ok(self.machine.our_value(lhs <= rhs)),
```
brson, line 45: Appling < and > to bools seems of questionable utility to me - that's why this match doesn't do it already. I didn't realize rust allowed these operators to be applied to bools. @nikomatsakis do you want < and > to apply to bools?

brson: I guess if it were to be used as keys in maps, then these operators would need to apply.

components/dada-execute/src/step/apply_op.rs
```diff
@@ -100,22 +106,34 @@ impl Stepper<'_> {
...
+                Op::LessEqual => {
+                    let val = lhs <= rhs;
+                    Ok(self.machine.our_value(val))
+                }
+                _ => op_error(),
+            },
             (&ObjectData::Unit(()), &ObjectData::Unit(())) => match op {
                 Op::EqualEqual => Ok(self.machine.our_value(true)),
+                Op::GreaterEqual => Ok(self.machine.our_value(lhs >= rhs)),
+                Op::LessEqual => Ok(self.machine.our_value(lhs <= rhs)),
```
brson, line 136: This again feels like it has little utility but I see that rust implements it. Maybe it's because there's a reasonable definition so why not? Maybe its to fulfill interfaces that might want ordering on units?

brson:
Patch looks great @Veeupup. My comments are mostly looking for insight from niko.

nikomatsakis:
Hey! Sorry that I haven't given any feedback! I had to take a bit of a break from Dada for a bit because "exhausted" but I'm psyched to get back to it. So let me read this PR! :)

components/dada-execute/src/step/apply_op.rs
nikomatsakis: I think it makes sense for the reason you said -- being usable in maps, or in sorted vectors, with binary search. It's often not bools on their own but rather some struct that contains a boolean field.

components/dada-execute/src/step/apply_op.rs
nikomatsakis: As above, in Rust terms, I think you want `Foo<()>` to be ordered...

```rust
struct Foo<T> { v: u32, extra: T }
```

components/dada-ir/src/code/validated/op.rs
```diff
@@ -37,6 +37,8 @@ impl std::fmt::Display for Op {
 
 define_operators! {
     EqualEqual => "==",
+    GreaterEqual => ">=",
+    LessEqual => "<=",
```
nikomatsakis, line 41: I am reminded that it would be really cool if the Python thing of `3 <= X <= 5` worked-- but that's independent from this PR.

nikomatsakis (approved):

PR #158, opened 2022-04-18
feature: `fn ->` syntax

### LINKED ISSUE #103: `fn ->` syntax
In #54, we decided that we should modify functions so that, to return a value, they have to be written with `->`. The idea then is that

```
fn foo() {
    22
}
```

in fact returns unit, but 

```
fn foo() -> {
    22
}
```

returns an integer.

This is to avoid accidentally returning the final value of a block.

In the future, `->` will be used to introduce a type annotation for the return value as well, just like Rust.

vemoo:
To parse it I added a new op `RightArrow` for `->`.
Then I validate that `return`s have a value if the function has a `->` and don't have it if it doesn't.
If the function doesn't have `->` I add a synthetic `{}` at the end on validation.

I'm not sure if this should be allowed:

```
fn foo() {
  ...
}

fn bar(x) {
  if x == 0 {
    return foo()
  }
  ...
}
```

With this PR it isn't and should be written as:

```
fn foo() {
  ...
}

fn bar(x) {
  if x == 0 {
    foo()
    return
  }
  ...
}
```

I have updated existing tests to use `->` where required and bless them.

There are a few changes to the diagrams code but I think those don't change the graph.

fixes #103


nikomatsakis:
Ooh, nice!

components/dada-ir/src/code.rs
```diff
@@ -14,6 +14,8 @@ pub struct Code {
     /// Tokens for the parameter list (parsed when we generate the syntax tree).
     pub parameter_tokens: Option<TokenTree>,
 
+    pub return_type: Option<FileSpan>,
```
nikomatsakis, line 17: can you add a comment explaining what this is the span of etc?

components/dada-validate/src/validate/validator.rs
```diff
@@ -150,7 +150,30 @@ impl<'me> Validator<'me> {
...
     #[tracing::instrument(level = "debug", skip_all)]
-    pub(crate) fn validate_expr(&mut self, expr: syntax::Expr) -> validated::Expr {
+    pub(crate) fn validate_root_expr(&mut self, expr: syntax::Expr) -> validated::Expr {
+        let validated_expr = self.validate_expr(expr);
+        if let Some(return_type) = self.code.return_type {
+            if let validated::ExprData::Seq(exprs) = validated_expr.data(self.tables) {
+                if exprs.is_empty() {
+                    dada_ir::error!(return_type, "must return something",)
+                        .primary_label("because return type is not unit")
+                        .emit(self.db);
```
nikomatsakis, line 160: Let's see if we can tweak this wording...

```suggestion
                    dada_ir::error!(return_type, "function body cannot be empty",)
                        .primary_label("because function is supposed to return something")
                        .emit(self.db);
```

Maybe this?

components/dada-validate/src/validate/validator.rs
```diff
@@ -520,12 +543,26 @@ impl<'me> Validator<'me> {
                 self.add(validated::ExprData::Seq(validated_exprs), expr)
             }
             syntax::ExprData::Return(with_value) => {
-                if let Some(return_expr) = with_value {
-                    let validated_return_expr = self.validate_expr(*return_expr);
-                    return self.add(validated::ExprData::Return(validated_return_expr), expr);
+                match (self.code.return_type, with_value) {
```
nikomatsakis, line 546: This is clever :) I was imagining we'd enforce this more dynamically, but I like the idea of checking it statically.

components/dada-validate/src/validate/validator.rs
```diff
@@ -520,12 +543,26 @@ impl<'me> Validator<'me> {
...
+                    (Some(return_type), None) => {
+                        dada_ir::error!(self.span(expr), "must return something")
+                            .primary_label("must return something")
+                            .secondary_label(return_type, "because return type is not unit")
+                            .emit(self.db);
```
nikomatsakis, line 551: Maybe this wording?

```suggestion
                        dada_ir::error!(self.span(expr), "return requires an expression")
                            .primary_label("cannot just have `return` without an expression afterwards")
                            .secondary_label(return_type, "because the function returns a value")
                            .emit(self.db);
```

Probably too wordy.

components/dada-validate/src/validate/validator.rs
```diff
@@ -520,12 +543,26 @@ impl<'me> Validator<'me> {
...
+                    }
+                    (None, Some(return_expr)) => {
+                        dada_ir::error!(self.span(*return_expr), "cannot return a value")
+                            .primary_label("cannot return a value")
+                            .emit(self.db);
```
nikomatsakis, line 556:
```suggestion
                        dada_ir::error!(self.span(*return_expr), "this function doesn't return anything")
                            .primary_label("this function can only `return`, without a value")
                            .emit(self.db);
```

Hmm, it doesn't have to be this PR, but I think it'd be good to show the user *where* the `->` *would* go. For this PR, can we add a label underlining the name and saying something like "because this function doesn't have a `->`"?

nikomatsakis (changes requested):
I left a few wording nits, though I don't think my wording is so great either. The biggest change I'd like to see, but it doesn't have to be in this PR, is storing the span of the *whitespace* where the `->` *would* go, so that we can print something like

```
Error: cannot return a value in this function
   ╭─[dada_tests/validate/return_type.dada:2:5]
   │
 1 │ fn not_unit1() {
   ·               ┬  
   ·               ╰── because function doesn't have `->` here
 2 │     return foo
   ·     ───┬──────  
   ·        ╰──── cannot return a value in this function
───╯
```

components/dada-validate/src/validate/validator.rs
vemoo: I'm thinking that can implement it by changing `return_type` to be `FileSpan` which would be the span of `->` or where it should go if it doesn't have it, and separately store a flag like `returns_value`. But I'm not sure it `Code` is the best place to store that information.

vemoo: Maybe a cleaner way of doing this is to have:
```rust
pub enum ReturnType {
    Value(FileSpan), // in the future this could be TokenTree?
    Unit(FileSpan),
}
```

Although seeing that `effect` in `Code` doesn't store `FileSpan` and requires using https://github.com/dada-lang/dada/blob/032f31282da9a9d73b095b8e1784fe99b5407b75/components/dada-validate/src/validate/validator.rs#L34 to be able to later get the span, I'm wondering if there's a reason to avoid storing `FileSpan` or `Filename` in `Code`.

nikomatsakis: I do try to avoid storing `FileSpan` because I want to get more incremental re-use. A good idea when you have something "spanned" is to create an entity for it -- 

e.g.

```rust
salsa::entity2! {
    entity ReturnType in crate::Jar {
        returns_value: bool,
        span: FileSpan,
    }
}
```

Not sure about using a `bool`, but you get the idea.

components/dada-ir/src/code.rs
```diff
@@ -14,6 +14,12 @@ pub struct Code {
     /// Tokens for the parameter list (parsed when we generate the syntax tree).
     pub parameter_tokens: Option<TokenTree>,
 
+    /// Span of `->` symbol if it has one, or where it would go if doesn't have it.
+    pub return_type: FileSpan,
```
nikomatsakis, line 18: See my other comment about why I try to avoid including a span directly in this structure -- in short, if the only thing that has changed is the span, I don't want to have to recompile.

nikomatsakis (changes requested):
Other than the nit, looks great!

nikomatsakis:
(needs rebase, `cargo dada test --bless`)

components/dada-ir/src/code.rs
vemoo: I have created `ReturnType` entity and also added a `ReturnTypeKind` enum to represent unit or value.

components/dada-validate/src/validate/validator.rs
vemoo: I have created the `ReturnType` entity and `ReturnTypeKind` enum with `Value` or `Unit`, which is a bit of a mouthful but better than a `bool`.

nikomatsakis (approved):
This is really nice work!

nikomatsakis:
I'm very excited about this, thanks @vemoo !

PR #161, opened 2022-04-20
speed up `dada test`

vemoo:
When implementing #158 I felt that running tests was taking too long. 
`perf` shows a lot of time spent in `cargo`, which happens because `test_dada_file_in_ide` spawns `cargo dada ide`.
But since `ide` functionality is in the same binary we can avoid going through `cargo`.

In my machine with this change running `cargo dada test` goes from 30s to 9s.

### DIFF
--- a/components/dada-lang/src/test_harness.rs
+++ b/components/dada-lang/src/test_harness.rs
@@ -33,6 +33,9 @@ impl Options {
             eyre::bail!("no test paths given; try --dada-path");
         }
 
+        let mut lsp_client = lsp_client::ChildSession::spawn();
+        lsp_client.send_init()?;
+
         const REF_EXTENSIONS: &[&str] = &["ref", "lsp", "bir", "validated", "syntax", "stdout"];
 
         for root in &self.dada_path {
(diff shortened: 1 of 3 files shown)


vemoo:
I see that on CI it runs `cargo test` which ends up executing the same tests but through `tests/harness.rs`. Which I think means that `components/dada-lang/src/test_harness.rs` gets compiled twice, once into `dada` binary and again into the `harness` integration test.

A way I see to avoid this issue and compiling twice is to invoke `dada test` from the `harness` integration test.

vemoo:
Another interesting thing is that with this changes when building in release mode, `dada test` takes 0.3s. Even with just with `opt-level = 1` it takes 0.5s.

@nikomatsakis should I wait for review or can I merge it with bors myself?

nikomatsakis:
@vemoo review from some other team member is good -- sorry for the delay, was very busy this week. Let me give this a try.

nikomatsakis (approved):
This is great, dada test is definitely slow. I have mixed feels about re-using the LSP process because it may mean that an individual test is failing because of accumulated error from other tests and not the test itself -- on the other hand, it's testing something kind of interesting on the side. Let's try it and see how it goes!

PR #162, opened 2022-05-01
Move webpage from mdbook to docusaurus

nikomatsakis:
This will allow in-place editing of examples and so forth! So cool.


XFFXFF:
I tried it locally, and it's really cool

nikomatsakis:
bors r- 

I'm suspicious this got stuck somehow, but I can't figure out how to see logs.

nikomatsakis:
Huh.

.github/workflows/test.yml
```diff
@@ -45,6 +45,24 @@ jobs:
         with:
           command: test
 
+  test:
```
XFFXFF, line 48: Is it because there are two tags called "test"?  
image

https://github.com/dada-lang/dada/actions/runs/2257562812/workflow

nikomatsakis:
@XFFXFF quite possible!

nikomatsakis:
bors r- 

grrr what the heck?

mbrobbel:
We need to update https://github.com/dada-lang/dada/blob/main/.github/bors.toml#L6 to reflect the update check name.

nikomatsakis:
@mbrobbel thanks! 🤞

PR #164, opened 2022-05-03
Change the layout of the playground

XFFXFF:
What does this pr do? 
* Change the layout to make the components fill the viewport more.
* Add a block for the state before the selected statement 
* Make the output display multi-line strings. 

The current playground seems to have some bugs, I can follow up and try to fix them.  

I don't have much experience with web front end, feel free to request changes.


nikomatsakis:
bors r+

this definitely looks better than what we have! :)

PR #166, opened 2022-05-03
Allow files to have "top-level" expressions in lieu of a main function

nikomatsakis:
Goal is to accept files like

```
print("Hello, world').await
```

Plan:

* [x] Refactor code to not expect a main `Function` per se
* [x] Parse top-level expressions and create a `SyntaxTree` with results
* [x] Validate top-level syntax tree and lower into `Bir` (probably requires creating a synthesized, temporary `Code`)
* [x] Alter `db.main_function(file)` to return the above
* [x] Write some tests! 
* [x] Write more tests! Fix bug in parser if it consumes just async!


nikomatsakis:
OK, I think this is ready

PR #168, opened 2022-05-03
playground: convert the output from the compiler into nice colors

### LINKED ISSUE #167: playground needs `ansi_up` integration
The old playground used the [`ansi-up`](https://github.com/drudru/ansi_up) library to convert the output from the compiler into nice colors.We ought to be able to do so again, but now that we are using docusaurus we can add it as a proper npm dependency I imagine by running

```bash
$ cd book; npm install --save ansi_up
```

and then invoking it in the proper places.

cc @XFFXFF, you mentioned interest in fixing playground bugs

XFFXFF:
fixes #167 

Nice colored output is back!
image

### DIFF
--- a/book/package-lock.json
+++ b/book/package-lock.json
@@ -12,6 +12,7 @@
         "@docusaurus/preset-classic": "2.0.0-beta.18",
         "@mdx-js/react": "^1.6.22",
         "ace-builds": "^1.4.14",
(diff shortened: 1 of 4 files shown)


nikomatsakis (approved):

nikomatsakis:
bors r+

nice!

PR #169, opened 2022-05-04
playground: make the borrow checker happy when editing

XFFXFF:
Now the playground throws a lot of errors saying "attempt to use a moved value" even if I just press "Enter" to wrap the line. This is because when we are editing, not only one event will occur, but at least one source change event and one cursor change event, which may cause concurrent access to `dada`.  In fact, we have designed a `Queue` to solve this, but we re-new a queue during every render, so I moved the queue declaration out of the function to keep it between re-renders

image

### DIFF
--- a/book/src/components/Ide/index.tsx
+++ b/book/src/components/Ide/index.tsx
@@ -101,9 +101,9 @@ class DCW {
 export type Cursor = { row: number; column: number };
 
 function Ide(props: { mini: boolean, sourceText: string }) {
-    const queue = new Queue();
     const [_module, setModule] = useState<InitOutput | null>(null);
     const [dada, setDada] = useState<DCW | null>(null);
+    const [queue] = useState<Queue>(() => new Queue());
 
     // First pass: we have to initialize the webassembly and "DCW"
     // instance.
--- a/book/src/pages/index.tsx
+++ b/book/src/pages/index.tsx
@@ -1,11 +1,9 @@
 import React from 'react';
 import clsx from 'clsx';
 import Layout from '@theme/Layout';
-import Link from '@docusaurus/Link';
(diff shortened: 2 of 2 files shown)


book/src/components/Ide/index.tsx
```diff
@@ -100,8 +100,9 @@ class DCW {
 
 export type Cursor = { row: number; column: number };
 
+const QUEUE = new Queue();
```
vemoo, line 103: Doing it like this means that all `Ide` components will use the same queue. A way to have each component have a different queue that is only created once is to do:
```
const [queue] = useState(() => new Queue());
```

Another option is to move the queue inside `DCW` which is already only created once.

XFFXFF: Yeah, it makes sense. 

I thought before that `useState` has to return a setter, I also learned how to do lazy initialization from your suggestion. Thanks!

nikomatsakis (approved):
Cool -- I didn't know about `useState(() => ...)` either.

book/src/components/Ide/index.tsx
vemoo: Well, `useState` returns an array with 2 elements, and `const [state, setState] = useState(0)` is using [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) to get those values, which is just syntactic sugar for:
```
const _xs = useState(0)
const state = _xs[0]
const setState = _xs[1]
```
so `const [state] = useState(0)` like:
```
const _xs = useState(0)
const state = _xs[0]
```

XFFXFF: @vemoo You are so sweet! Thanks a lot for the extra explanation 💖

PR #170, opened 2022-05-06
get `cargo xtask deploy` working on mac

nikomatsakis:
The current `cargo xtask deploy` script assumes linux. I'm trying to get it to run on mac (and hopefully windows). Not there yet.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,42 +1,6 @@
 Dada is pretty young and informal, but we welcome contributions! 
 
-Keep in mind that Dada is my "spare time, relaxation and exploration" project. Responses may be slow and I am driving the language direction, syntax, and semantics. Naturally I would absolutely love to hear what people think, but I am acting as Chief Dadaist and final decision maker for the time being. --nikomatsakis
+You can read more about contributing here:
 
-## Feature gates
+https://dada-lang.org/docs/contributing/guidelines
 
-Dada uses nightly, but we are careful with the feature gates we use. We only want to use features
-that are close to stabilizing. Currently accepted feature gates:
-
-```rust
-#![feature(let_else)]
-#![feature(trait_upcasting)]
-#![feature(try_blocks)]
-```
-
-## Finding an issue
-
-Issues on the repository that are tagged with "good first issue' and the like are a good bet.
-
-## Review policy
-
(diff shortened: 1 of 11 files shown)


nikomatsakis:
Current output on mac

```
error: failed to build archive: section too large

The following warnings were emitted during compilation:

warning: warning: /Library/Developer/CommandLineTools/usr/bin/ranlib: archive library: /Users/nikomat/versioned/dada/target/wasm32-unknown-unknown/debug/build/psm-caa62c7e694abb40/out/libpsm_s.a the table of contents is empty (no object file members in the library define global symbols)

error: could not compile `psm` due to previous error
```

Hey @dada-lang/dadaists, any tips? @brson I know you've spent some time looking at wasm errors before.

vemoo:
I cannot help with mac, but I can help with windows.

nikomatsakis:
@vemoo does `cargo xtask deploy` work on windows? I have a windows computer but I haven't given it a try...

vemoo:
`wasm-pack` fails because of [this issue](https://github.com/rustwasm/wasm-pack/issues/1097), I can work around it by changing the version to `v0.10.1`.

But then I get:
```
[INFO]: Compiling to Wasm...
...
error: failed to run custom build command for `psm v0.1.16`
...
error occurred: Failed to find tool. Is `ar` installed? (see https://github.com/alexcrichton/cc-rs#compile-time-requirements for help)
```
but I don't think `psm` can work on wasm so maybe the functionality that depends on it can be conditionally disabled when building for wasm.

vemoo:
It works if I remove `stacker` from `dada-parse` so it could be conditionally disabled when the target is wasm.

Then there's `` Error: command not found: `npm` ``, which can be worked around by using `npm.cmd` instead.

And then it works.

nikomatsakis:
We're using stacker? that is probably overkill....

nikomatsakis:
OK, I just removed stacker for now and that seemed to get it working on mac. 🎉 

We could always add it back later but I'm not especially worried about stack overflow right now.

I added a "Contributing" section to the website that documents how to build things.

nikomatsakis:
It sounds like windows needs a bit more work, I'm going to change the subject line of this PR -- thanks for investigating, @vemoo !

PR #171, opened 2022-05-07
parse parenthesized expressions

### LINKED ISSUE #141: `(expr)` is not supported
Things like `(1)` or `(1 + 1)` will get an error saying  
```
thread 'main' panicked at 'assertion failed: `(left != right)`
  left: `1`,
 right: `1`', components/dada-brew/src/brew.rs:334:25
```

related code   
https://github.com/dada-lang/dada/blob/5b5e366465c2bddec7db832a6e6cd6a2b4865c64/components/dada-brew/src/brew.rs#L323-L339

vemoo:
As per doc comments tuples are of `len != 1` https://github.com/dada-lang/dada/blob/0793cdf392f8028b7f00e63cef304e51b77f7da5/components/dada-ir/src/code/syntax.rs#L135-L136

so I fixed parsing to match that.

fixes #141


vemoo:
```dada
print((1))
```
is causing this assert to fail: https://github.com/dada-lang/dada/blob/0793cdf392f8028b7f00e63cef304e51b77f7da5/components/dada-validate/src/validate/validator.rs#L205-L207

I have fixed it by doing:
```rust
syntax::ExprData::Parenthesized(parenthesized_expr) => {
    let data = self
        .validate_expr_in_mode(*parenthesized_expr, mode)
        .data(self.tables)
        .clone();
    self.add(data, expr)
}
```
in https://github.com/dada-lang/dada/blob/0793cdf392f8028b7f00e63cef304e51b77f7da5/components/dada-validate/src/validate/validator.rs#L400-L402

but I don't know if that's the proper fix.

vemoo:
Ok, I think I found the proper fix, by adding `validated::ExprData::Parenthesized`

components/dada-ir/src/code/syntax.rs
```diff
@@ -129,10 +129,10 @@ pub enum ExprData {
     /// `[shared|var|atomic] x = expr`
     Var(LocalVariableDecl, Expr),
 
-    /// `expr`
+    /// `(expr)`
     Parenthesized(Expr),
```
XFFXFF, line 133: We don't seem to have anything parsed as a `Parenthsized` before 😅  .

components/dada-ir/src/code/validated.rs
```diff
@@ -258,6 +258,9 @@ pub enum ExprData {
     /// `expr.give`
     Give(Place),
 
+    /// `(expr)`
+    Parenthesized(Expr),
```
nikomatsakis, line 262: Hmm, I think we shouldn't have this in `Validated`....

nikomatsakis: ...it belongs in `Syntax`, but `Validated` is supposed to be kind of only the things that have meaning to the compiler

nikomatsakis (changes requested):
Nice! I think that the assertion is probably just wrong, though,.

components/dada-validate/src/validate/validator.rs
```diff
@@ -398,7 +398,8 @@ impl<'me> Validator<'me> {
             }
 
             syntax::ExprData::Parenthesized(parenthesized_expr) => {
-                self.validate_expr_in_mode(*parenthesized_expr, mode)
+                let validated_expr = self.validate_expr_in_mode(*parenthesized_expr, mode);
+                self.add(validated::ExprData::Parenthesized(validated_expr), expr)
```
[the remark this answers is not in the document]
vemoo: I had to change that because otherwise this assert was failing: https://github.com/dada-lang/dada/blob/60986a30ec5ccf130318335231ff4aabe28904c9/components/dada-validate/src/validate/validator.rs#L194-L196

Before adding `Parenthesized` to `Validated` I did https://github.com/dada-lang/dada/blob/e4c5bdc4866a17666206b9544a8b84edc2fa278a/components/dada-validate/src/validate/validator.rs#L400-L406
but that felt wrong. ¿Maybe that is a better fix?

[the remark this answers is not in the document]
vemoo: Ah, you mean the assertion in validator, not the one from the #141, right?
There's another one in `reserve_validated_expr`

components/dada-parse/src/parser/code.rs
```diff
@@ -449,7 +449,15 @@ impl CodeParser<'_, '_> {
         } else if let Some((span, token_tree)) = self.delimited('(') {
             let expr =
                 self.with_sub_parser(token_tree, |subparser| subparser.parse_only_expr_seq());
-            Some(self.add(ExprData::Tuple(expr), span))
+
+            Some(self.add(
```
nikomatsakis, line 453: For better or worse, I think this change will mean that `(a,)` also parses as a parenthesized expression. We should check that. Not sure if we want that or not!

nikomatsakis (changes requested):
Looked it over, this seems better, did have one thought...

components/dada-parse/src/parser/code.rs
vemoo: I will investigate the best way to emit an error for that case.

nikomatsakis: Probably we want an error? It's not entirely clear to me.

nikomatsakis: I'd be ok with just opening up an issue to follow-up on later.

vemoo: created #177

nikomatsakis (approved):

PR #172, opened 2022-05-10
Format strings

### LINKED ISSUE #31: format strings
I didn't fully implement format strings. This is what I had in mind:

`foo{bar}baz` compiles down to:

* a syntax tree with `foo{bar}baz`
* a validated tree with `{ let x = bar; Concatenate("foo", x, "baz") }` where `Concatenate` is a new [intrinsic](https://github.com/dada-lang/dada/blob/main/components/dada-ir/src/intrinsic.rs)
* at that point, later phases don't have to know anything about it

Right now we special case format strings without code in the parser:

https://github.com/dada-lang/dada/blob/363cc82ce9d67bc0e921a303770762dd4eb6d152/components/dada-parse/src/parser/code.rs#L366-L379

we should stop doing that, and instead modify the syntax definition so that `StringLiteral` becomes `FormatString` and carries a [`FormatString`](https://github.com/dada-lang/dada/blob/main/components/dada-ir/src/format_string.rs):

https://github.com/dada-lang/dada/blob/363cc82ce9d67bc0e921a303770762dd4eb6d152/components/dada-ir/src/code/syntax.rs#L101-L105

then we would modify the validated tree code to create the intrinsic:

https://github.com/dada-lang/dada/blob/363cc82ce9d67bc0e921a303770762dd4eb6d152/components/dada-validate/src/validate/validator.rs#L164-L166

vaguely like we desguar `x += y` to `x = x + y`:

https://github.com/dada-lang/dada/blob/363cc82ce9d67bc0e921a303770762dd4eb6d152/components/dada-validate/src/validate/validator.rs#L347-L365

nikomatsakis:
Implement format strings, closing one of the remaining gaps that makes the tutorial not directly usable

Closes #31


nikomatsakis:
Hey @dada-lang/dadaists, if one of you has time, perhaps give this a quick review? I've just been r+'ing my own, mostly because I was eager to see it show up on dada-lang.org :grin: but it seems better if every PR gets reviewed.

nikomatsakis:
Playing on the playground, I found a bug:

```
print("
    I have forced {"myself"} to contradict myself
    in order to avoid conforming to my own taste.
        -- Marcel Duchamp
").await
```

I think the problem is that whitespace stripping doesn't necessarily interact well

XFFXFF (approved):
Cool! It looks good to me.  
I remember I tried to fix this issue before and it failed. It's a really good experience to see how you solve it.  
Actually, I have read your previous PRs, I'm still learning the source code and compiler, sorry for not being able to give you some constructive review comments.

vemoo (approved):

nikomatsakis:
bors r+ 

Thanks y'all!

PR #178, opened 2022-05-12
fix margin computation with format strings

### LINKED ISSUE #174: whitespace stripping and format strings don't play well
This code prints "forcedmyselfto" and not "forced myself to" as you might expect

```
print("
    I have forced {"myself"} to contradict myself
    in order to avoid conforming to my own taste.
        -- Marcel Duchamp
").await
```

_Originally posted by @nikomatsakis in https://github.com/dada-lang/dada/issues/172#issuecomment-1122233825_

nikomatsakis:
Fixes #174


dada_tests/interpret/string-escape/stdout.ref
```diff
@@ -8,9 +8,11 @@ Foo " Bar
 Foo \
 	    Foo
     	Bar
+
```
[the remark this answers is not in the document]
nikomatsakis: and probably add `#! OUTPUT` statements to the tests

components/dada-validate/src/validate/validator.rs
```diff
@@ -767,6 +760,124 @@ impl<'me> Validator<'me> {
...
+                //
+                // and we want that to print "Foo". Note that we haven't applied
+                // escapes yet, so `print("\n")` is unaffected.
+                if index == 0 {
+                    if let Some(s) = without_margin.strip_prefix('\n') {
+                        without_margin = s;
+                    }
+                } else if index == exprs.len() - 1 {
+                    if let Some(s) = without_margin.strip_suffix('\n') {
+                        without_margin = s;
```
XFFXFF, line 795: When will we get here? I can't find a case. I think the suffix `\n` is always striped in `strip_margin`.

XFFXFF (commented):
Nice!

components/dada-validate/src/validate/validator.rs
XFFXFF: Oh, I found a case
```
async fn main() {
    print("
    hello
    {"world"}
    ").await
}
```
But it does not strip trailing newline for a string literal, like the documented example, because the `index` is 0 and we don't get into this condition branch.
```
async fn main() {
  print("
  foo
  ").await
}
```

components/dada-validate/src/validate/validator.rs
```diff
@@ -767,6 +760,124 @@ impl<'me> Validator<'me> {
...
+
+        self.add(
+            validated::ExprData::Concatenate(validated_exprs),
+            concatenate_expr,
+        )
+    }
+
+    /// Returns the number of characters that should be stripped from the start of
+    /// every (non-empty) line. Dada format string literals automatically string
+    /// the "margin", which is common whitespace that appears at the start of
```
XFFXFF, line 816: A typo, string the margin -> strip the margin

components/dada-validate/src/validate/validator.rs
nikomatsakis: Hmm, yeah, something is broken with this logic

nikomatsakis:
@XFFXFF ok, I took another pass at this this morning and spent altogether too much time on it. =) Take a look, I think it makes sense now. I also wrote up some reference material. I also cleaned up the escapes a bit as part of that story.

The one thing that changed behavior with the new code is the test that consists of: all blank lines but which have trailing whitespace. That trailing whitespace is now preserved, because the margin is computed as 0, since the margin computation ignores blank lines. I am ok with that, I think, though I could also see the logic in removing all trailing whitespace from each line -- or perhaps only ignoring lines that are truly empty, with *no* trailing whitespace. Perhaps that would be the most consistent overall -- either trailing whitespace matters or it doesn't.

(That reminds me that I think `\ ` would be a useful escape, but I didn't add that here.)

XFFXFF (approved):
Super cool!  

Everything makes sense to me now. The reference is really helpful, I spent quite some time recalling the logic of strip during my previous review 😜

PR #183, opened 2022-05-16
playground: set new line mode to unix

### LINKED ISSUE #182: Heap graph is not displayed correctly in playground
The actual heap graph displayed at the cursor

image

Expect 

image

XFFXFF:
fixes #182


XFFXFF:
Yeah, I think #186 is a better solution. I will close this when #186 lands.

PR #184, opened 2022-05-16
enable typescript strict options, and run typechecking on deploy

vemoo:
There was an small issue already catched by typechecking, the `mode` property not being passed to the `Output` component in one place.

Also removed `remove-blank-lines` and `strip-indent` because they were missing type declarations, and they are only used on a static string. If later needed I think they are simple enough that it would be best to implement them ourselves.

I also added the typechecking step to `cargo xtask deploy`.


XFFXFF (approved):
Nice!

PR #185, opened 2022-05-16
fix ace theme and mode loading

vemoo:
Ace theme wasn't loading before, it seems the issue was that the mode needed to be set.

I chose rust mode, because it seems to have the most similarities for now.

Ace editor now also respects dark mode:

image


XFFXFF (approved):
Nice!  
Playground in dark mode looks much better :)

PR #186, opened 2022-05-16
offset to line column conversion fixes

vemoo:
Fixes issue detected in #182 and also another issue with multi-byte characters.

It's based on what rust-analyzer does: https://github.com/rust-lang/rust-analyzer/blob/825ce48180bb9344a198df9361bd43a40b5b13d4/crates/ide-db/src/line_index.rs

### DIFF
--- a/components/dada-ir/src/lines.rs
+++ b/components/dada-ir/src/lines.rs
@@ -5,42 +5,32 @@ use crate::{
 
 #[derive(Debug, PartialEq, Eq, Hash)]
 pub struct LineTable {
-    /// Stores the index of the `\n` for each line.
-    /// So `0..line_endings[0]` represents the range of characters for the first line
-    /// and so forth.
-    line_endings: Vec<Offset>,
+    /// Stores the index of the start for each line.
+    /// Always has Offset(0) at position 0.
+    line_starts: Vec<Offset>,
     end_offset: Offset,
 }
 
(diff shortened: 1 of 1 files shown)


components/dada-ir/src/lines.rs
```diff
@@ -1,70 +1,144 @@
...
-    };
-    for line in source_text.lines() {
-        p += line.len();
-        table.line_endings.push(Offset::from(p));
-        p += 1;
+    LineTable::new(source_text)
+}
+
+#[cfg(test)]
+mod tests {
```
XFFXFF, line 105: Maybe we should use our own test harness, like https://github.com/dada-lang/dada/blob/main/dada_tests/heap-graph/cursor-position.dada ?  But I think the tests you  wrote here are more intuitive.

XFFXFF (commented):
Nice! This is a more general solution than I did in #183. I like it

components/dada-ir/src/lines.rs
vemoo: The way I see it this are unit tests and the ones from the test harness are integration tests, so I think they are complementary.

nikomatsakis: Agreed, this seems like a classic and perfect use for unit tests

nikomatsakis (approved):

PR #190, opened 2022-05-29
Some small styling improvements

ciyer:
* add some space on the left of the "Dada Playground" header and control elements
* clicking the model label switches modes
* set cursor over mode label to indicate that it is clickable
* use background color and left border as a separator

Light Mode
image

Dark Mode
image


vemoo:
It doesn't work properly with dark mode.
Before:
dada-lang org_

After:
deploy-preview-190--dada-lang netlify app_playground

ciyer:
Oh, that looks terrible, sorry about that! I have fixed the CSS to use the color variables defined in the themes.

Dark mode now looks like this.

image

vemoo (approved):
Nice!

PR #191, opened 2022-06-02
fix md and avoid backticks coexist with embed expr

liigo:
make it clear that backticks have nothing to do with the embed expr grammar


nikomatsakis (approved):

nikomatsakis:
makes sense, thanks!

PR #192, opened 2022-06-26
make "in page" editor be less terrible

nikomatsakis:
still pretty terrible, but less terrible


nikomatsakis:
Apparently my auto-formatter is set to 2-space indent, and whoever last modified these files had different settings. Not even sure how I would configure that, but seems like we should settle on *something*!

nikomatsakis:
Hmm, it seems I somehow broke the `syntax` viewing and all that

nikomatsakis:
> But maybe we could setup [prettier](https://prettier.io/) to handle formatting?

that would be ideal, yes

Issue #193, opened 2022-07-02
whitespaces are lost in html generated from docs/reference/string-literals.md


vemoo:
In the source file [string-literals.md](https://github.com/dada-lang/dada/blob/eeff91e25a9baaaf44beca0ee6e5d3172e33fd2e/book/docs/reference/string-literals.md?plain=1) there are a couple of places that have multiple consecutive white spaces. For example:
```
This is equivalent to "\n    Hello,\n      world\n". The initial `{""}` disables all subsequent margin stripping.
```
but those are lost after converting it to html:

> This is equivalent to "\n    Hello,\n      world\n". The initial `{""}` disables all subsequent margin stripping.

PR #194, opened 2022-07-02
setup prettier

vemoo:
I have left prettier default settings, but I'm open to suggestions.

Prettier can also format *.md files, but I have excluded them for now because it removes the whitespaces I mention in #193.


XFFXFF (approved):
nice!  
But clippy looks a little unhappy :stuck_out_tongue_winking_eye:

PR #198, opened 2022-07-14
fix tutorial link in README

ammkrn:
The current tutorial link returns a page not found error.


lqd:
good catch, thank you!

bors r+

PR #202, opened 2022-08-01
Parse and validate break/continue

### LINKED ISSUE #127: break/continue are not parsed
I just noticed that `break` and `continue` are not parsed. =) I think the validation/BIR code is mostly there. We should fix that!

### Expected syntax

```
break <expr>
continue
```

We should probably add some syntax for named loops but that seems like a separate, follow-on feature.

### Precedence

The precedence of `break` and `continue` should be *very* low, e.g., `break 22 + 44` should yield break with `66`, and `break foo().await` should do the expected thing.

Aadamandersson:
This adds support for break/continue and closes #127.


components/dada-validate/src/validate/validator.rs
```diff
@@ -515,6 +515,40 @@ impl<'me> Validator<'me> {
                     exprs.iter().map(|expr| self.validate_expr(*expr)).collect();
                 self.add(validated::ExprData::Seq(validated_exprs), expr)
             }
+            syntax::ExprData::Continue => {
+                let validated_data = match self.loop_stack.pop() {
```
nikomatsakis, line 519: Do we really want to *pop* from the stack? I would guess "peek".

Maybe add a test like this:

```
loop {
    if false { continue }
    if false { continue }
}
```

and see what happens

components/dada-validate/src/validate/validator.rs
```diff
@@ -515,6 +515,40 @@ impl<'me> Validator<'me> {
...
+                        dada_ir::error!(self.span(expr), "cannot `continue` outside of a loop")
+                            .primary_label("`continue` outside of a loop here")
+                            .emit(self.db);
+                        validated::ExprData::Error
+                    }
+                };
+                self.add(validated_data, expr)
+            }
+            syntax::ExprData::Break(with_value) => {
+                let validated_data = match self.loop_stack.pop() {
```
nikomatsakis, line 531: as above, 

```
loop {
    if false { break }
    if false { break }
}
```

will break here, I think

nikomatsakis (changes requested):
Cool! I left one nit.

nikomatsakis:
Actually, can you add a test for that last bug fix?

nikomatsakis:
Oh, I see you're saying the test didn't fail as expected. Well, ok. Let's just land it.

bors r+

PR #203, opened 2022-08-02
update various packages, notably salsa

nikomatsakis:
new version of salsa is on master, but required significant changes.

ariadne seems to have had a bugfix.


components/dada-parse/src/parameter_parser.rs
```diff
@@ -3,7 +3,7 @@ use crate::parser::Parser;
 use dada_ir::{class::Class, function::Function, parameter::Parameter, token_tree::TokenTree};
 use salsa::DebugWithDb;
 
-#[salsa::component(in crate::Jar ref)]
+#[salsa::component(jar = crate::Jar, return_ref)]
```
XFFXFF, line 6: What's `salsa::component` for? can't find it in salsa's document.  Can we use `salsa::memoized` here?

XFFXFF (approved):
Nice! salsa's new api is cleaner.  
have a question ↓

components/dada-parse/src/parameter_parser.rs
nikomatsakis: Good question -- this has been renamed to the `specify` option for `salsa::tracked`, see [this part of the overview](https://salsa-rs.github.io/salsa/overview.html#specified-the-result-of-tracked-functions-for-particular-structs) for details.

Issue #17, opened 2022-01-08
Dada reference grammar


nikomatsakis:
Our parser is hand-written and I'm ok with that, but it'd be nice to have a reference grammar, particularly one that enforces non-ambiguity. 

I was thinking we could do it using some kind of parser generator tool that supports LL(k) or perhaps (better) GLL/GLR. Looking around at what I see on crates.io, parol looks like the best fit, and most actively maintained.

I definitely want something on crates.io so that we can make a `dada-reference-grammar` crate, run `cargo test` on it, and so forth, without having to install any external tools.

brson:
I've started on this.

brson:
parol has a neat feature that it can emit random valid syntax, which could be useful for testing.

brson:
Now it's only 9 tests failing, all related to format strings (which I haven't tried to parse yet); and return without value and unit/tuples. The latter two both seem ambiguous to me.

Return without expression because either:

- either the expression is optional ,r
- it is not possible to have any expression-statements after a return statement

e.g.

```
fn foo() {
    return
    dead_code_here // can't write this
}
```

Unit/tuples, at least in statement position, are ambiguous with function calls per https://github.com/dada-lang/dada/issues/117

brson:
I am though very happy for you to correct me and fix my grammar definition. I don't understand this subject well.

nikomatsakis:
Nice! I'll make some time to check this out.

brson:
I added two xtasks for grammar so you don't have to type the full commands:

- `cargo xtask grammar build`
- `cargo xtask grammar decidable` which shows the maximum lookahead

brson:
33 tests failing now. I'll give it another look soon.

brson:
Well I fixed the problem with return-without-expr by making it only possible to appear as the final expression in a block.

brson:
If we're going to be newline-sensitive for disambiguation elsewhere then we should probably do it for return/break/continue as well.

brson:
Unasigning myself. Busy with other things lately.

vemoo:
Regarding
```
fn foo() {
    return
    bar
}
```
the current parser parses it as return with value `bar`, which I think it makes sense, because you can do:
```
fn foo() {
    return,
    bar
}
```
if you want a return without value.

vemoo:
I've been experimenting with tree-sitter here: https://github.com/vemoo/dada/blob/tree-sitter/tree-sitter-dada/grammar.js

[...]

PR #204, opened 2022-08-03
add source to diagnostics

vemoo:
This will help with #17, by making it easier to know for which files the reference parser should emit an error.

We could also complete the expected diagnostics in tests to include the source, but I'm not sure if that is worth it. At the moment we only distinguish between runtime and compile-time diagnostics.

### DIFF
--- a/components/dada-ir/src/diagnostic.rs
+++ b/components/dada-ir/src/diagnostic.rs
@@ -13,6 +13,9 @@ pub struct ErrorReported;
 #[derive(Clone, PartialEq, Eq, Hash, Debug)]
 #[non_exhaustive]
 pub struct Diagnostic {
+    /// This is only used for testing the reference grammar.
+    /// If it's true it means that the reference grammar should also emit an error.
+    pub is_grammar_error: bool,
     pub severity: Severity,
     pub span: FileSpan,
     pub message: String,
@@ -51,6 +54,15 @@ macro_rules! diag {
 macro_rules! error {
     ($span:expr, $($message:tt)*) => {
         $crate::diagnostic::Diagnostic::builder($crate::diagnostic::Severity::Error, $span, format!($($message)*))
+    }
+}
+
+/// Convenience macro for avoiding `format!`
+/// Use this when the error should also be emitted by the reference grammar
+#[macro_export]
+macro_rules! grammar_error {
+    ($span:expr, $($message:tt)*) => {
+        $crate::diagnostic::Diagnostic::builder($crate::diagnostic::Severity::Error, $span, format!($($message)*)).is_grammar_error()
     }
 }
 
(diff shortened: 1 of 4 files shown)


nikomatsakis:
@vemoo -- I'm not sure how this helps, can you explain a bit more? 

(Also, it needs a rebase, if nothing else, sorry about that...)

vemoo:
To verify the tree-sitter grammar I added this: https://github.com/vemoo/dada/blob/af9cdc0c61cb6de421c733fd73fa167ed6ac716b/components/dada-lang/src/test_harness.rs#L179-L199

But I don't think the grammar should catch validation errors, so we need to know which ones are from lexing or parsing.

I see that brson here: https://github.com/brson/dada/blob/e2861cc6b0c00af2a4e52aa7d6ad93813ad58818/components/dada-lang/src/test_harness.rs#L519-L549 is comparing against the expected diagnostics. So maybe another option is to juste add something like `#! PARSE ERROR`?

nikomatsakis:
Ah, I see. So the idea is that we (internally) track the phase of the compiler the error comes from, for the purpose of our testing harness? In particular, we want to know whether it got a parse error, so that we can compare that against the reference grammar?

nikomatsakis:
What if instead we just distinguish parse error vs other error, and we make a separate macro `parse_error` or something like that? This patch seems kind of noisy and makes reporting errors more confusing. I feel like sometimes errors may change whether they are detected at validation or not and it'd be annoying to wonder if this will affect any tooling (same is true of parse errors, but they seem less likely to change). It'd also be good to document clearly the reason for the `is_parse_error: bool` flag on diagnostics and its role in testing.

vemoo:
Maybe instead of `parse_error` it makes more sense to call it `grammar_error`? Because it will also include lexing errors.

vemoo:
With this changes maybe it doesn't make sense to merge it as a separate PR and it should be part of the reference grammar PR?

nikomatsakis:
@vemoo I was just thinking -- another option is, rather than modifying the `Diagnostic` struct, it's possible to ask for *only the errors that resulted from parsing a given file*. 

The code for getting diagnostics today is:

https://github.com/dada-lang/dada/blob/0199d1bef32a259b30f5b5b721a89ca5cf16dff4/components/dada-db/src/lib.rs#L55-L60

but we could add another method:

```rust
    /// Checks `input_file` for parsing errors and returns relevant diagnostics.
    pub fn grammar_diagnostics(&self, input_file: InputFile) -> Vec<Diagnostic> {
        dada_parse::file_parser::parse_file::<dada_ir::diagnostic::Diagnostics>(
            self, input_file,
        )
    }
```

and it would just the diagnostics emitted by the parser. Maybe that'd be easier still?

vemoo:
Yeah, I like that. It ended up being a little more complicated: https://github.com/vemoo/dada/blob/29b3718387fe2ae3d4e5498455165d24ac1087b2/components/dada-parse/src/lib.rs#L26-L50 because `parse_file` leaves some parts unparsed as `TokenTree`s.

I'm going to close this PR because this change will make more sense in the context of adding a reference grammar.

PR #205, opened 2022-08-09
Tuple indexing expression support

Aadamandersson:
We do currently not have support for accessing the fields of a tuple, which I've added support for in this PR.

The syntax for a _tuple indexing expression_ is the same as in rust, e.g.,
```
async fn main() {
    t = ("a", "b")
    print("{t.0}").await #! OUTPUT a
    print("{t.1}").await #! OUTPUT b
}
```
I'm not sure if this is the syntax we want, but I'm happy to do something else as well.


nikomatsakis:
@Aadamandersson this is cool, thanks! Let me check it out.

components/dada-execute/src/step/traversal.rs
```diff
@@ -240,6 +242,15 @@ impl Stepper<'_> {
                     ))
                 }
             }
+            ObjectData::Tuple(tuple) => {
+                let field_name_str = field_name.as_str(self.db);
```
nikomatsakis, line 246: Nit: Feels like it might be better to parse the field name once into an integer...? Not that this code is mega optimized. :)

nikomatsakis: Oh, I see from your tests, maybe we don't want `foo.01` to work... hmm.... interesting question. :)

dada_tests/interpret/invalid_tuple_index.dada
```diff
@@ -0,0 +1,4 @@
+async fn main() {
+    tuple = ("a", "b", 3)
+    print("{tuple.01}").await #! RUN ERROR no field named `01`
+}
```
nikomatsakis, line 4: Another relevant example:

* `tuple.1_` -- I think this parses as a number

nikomatsakis (changes requested):
Can we add a test for nested tuples, like

```
let x = ((1, 2), 3);
print("{x.0.0}").await #! OUTPUT 1
print("{x.0.1}").await #! OUTPUT 2
print("{x.1}").await #! OUTPUT 3
```

nikomatsakis (approved):
Nice!

PR #209, opened 2022-08-23
Parse `else if`

Aadamandersson:
This adds support for parsing `else if`, which we probably want to have!

```
async fn main() {
    if false {
        print(1).await
    } else if true {
        print(2).await
        #! OUTPUT 2
   }
}
```


nikomatsakis (approved):
Good point :)

PR #210, opened 2022-09-22
refactor fn signatures to have syntax nodes

nikomatsakis:
I'm working towards a few goals right now, one of which is implementing optional type signatures in functions and local variables (e.g., `fn foo(x: my String)` or whatever). To do that, we are going to need to extend the syntax tree with additional nodes to represent types (the current setup, where `Ty` is an interned struct, doesn't really make sense, I realize). This branch takes a step towards that by creating a `syntax::Signature` struct and adding in various bits of the fn signature into there. 

I've also started a refactoring of how we manage spans, moving towards the "anchored" model demonstrated in https://github.com/salsa-rs/salsa/pull/396. This PR only takes a micro-step, removing the use of `FullSpan` and instead just using `Span` (but spans still contain absolute spans relative to the start of the file, not spans relative to the start of the function).

Writing all this out, I realize I should open up some issues for various refactorings and things  I have in mind, there's a couple different directions happening at once...


components/dada-execute/src/heap_graph/graphviz.rs
```diff
@@ -347,7 +347,7 @@ impl HeapGraph {
         if let Some(name) = name {
             w.permissions
                 .entry(edge.permission)
-                .or_insert(vec![])
+                .or_insert_with(|| vec![])
```
lqd, line 350:
```suggestion
                .or_default()
```

should fix it

nikomatsakis: see rust-lang/rust-clippy#9522 :)

Issue #217, opened 2022-10-03
Syntax question: to `let` or not to `let`


nikomatsakis:
In the current dada syntax, if you use `=` you declare a new variable:

```
x = 22
```

and if you use `:=`, you can reassign an existing place:

```
x := 23
y.field := 23
```

I was going for an extremely "lightweight" syntax reminiscent of Python, but to be honest, I'm not sure I like it. I am thinking maybe it's better to follow Rust (and, I realize, Java/TypeScript?) precedent here and use `let`:

```
let x = 22
```

and then just use `=` for reassignment:

```
x = 23
y.field = 23
```

Thoughts?

nikomatsakis:
This probably makes parsing easier too, as you would presumably write

```
let x: Type
```

and 

```
let atomic x
```

Right now we need a bit of goofy lookahead to handle these cases.

vemoo:
I prefer having `let`, it's easier to see where new bindings are being introduced.

nikomatsakis:
@vemoo want to try implementing it?

vemoo:
sure

brson:
I also prefer keywords to introduce variables. I think it's because it's a strong visual indicator that new state is being created.

vemoo:
this was implemented in #221

PR #221, opened 2022-10-06
Let declarations

vemoo:
Implements #217

I was going to update the book but I think it was already outdated since it uses `my foo = bar` instead of `foo: my = bar`. ¿Should updating the book be a separate PR?.

Also I didn't find any test for atomic local variable declarations, ¿should one be added?

### DIFF
--- a/components/dada-parse/src/parser/expr.rs
+++ b/components/dada-parse/src/parser/expr.rs
@@ -139,7 +139,7 @@ impl CodeParser<'_, '_> {
                     Op::MinusEqual,
                     Op::DividedByEqual,
                     Op::TimesEqual,
-                    Op::ColonEqual,
+                    Op::Equal,
                 ],
                 Self::parse_expr_5,
             ) {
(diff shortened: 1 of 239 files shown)


components/dada-parse/src/parser/expr.rs
```diff
@@ -359,22 +359,15 @@ impl CodeParser<'_, '_> {
...
-            this.parse_name().is_some()
-                && (this.eat_op(Op::Colon).is_some() || this.eat_op(Op::Equal).is_some())
-        })?;
-
-        // Look for `[mode] x = `. If we see that, we are committed to this
-        // being a local variable declaration. Otherwise, we roll fully back.
+        let (let_span, _) = self.eat(Keyword::Let)?;
+
         let atomic = self.parse_atomic();
         let name = self.parse_name().unwrap();
```
nikomatsakis, line 368: we need to report an error here -- e.g., people could write `let = 5`.

nikomatsakis (changes requested):
This is definitely a good change. Thanks for prep'ing the PR!

Regarding the book -- If the edits are small, I think it's fine to do in this PR, but I'm also ok with leaving it for a follow-up PR. I have vague plans to do another pass over the tutorial anyway as the type work is progressing.

components/dada-parse/src/parser/expr.rs
vemoo: For `let = 5`, does it make sense to introduce some kind of dummy empty `Name` in the ast between `let` and `=`, like for the value with `or_dummy_expr`?

This way an error will be emitted but it will be parsed. Otherwise `unexpected token` will also be emitted later for `let`.

vemoo: nevermind, I think I found a better solution, parse the whole declaration into an `Expr::Error`

vemoo:
Mm, I see now that `: our` alone is not valid syntax, and the type is required. So it seems that there's no equivalent for what `class Value(our n)` meant before, is there?

I didn't catch it before because it seems the playground is not showing all diagnostics.

nikomatsakis:
bors r+

No, there is no equivalent to `our x` from before. I'll go over the book edits later I think, it needs to be reworked anyhow.

PR #224, opened 2023-05-09
Dada Language Server Extension

### LINKED ISSUE #5: we need a vscode extension
There is an LSP server (dada-lsp) and we even test it, but we don't have an extension. I have no idea how to write one. =) The vscode server is pretty bare-bones but it should support opening/closing files and reporting errors.

Aadamandersson:
This implements a language server extension for Visual Studio Code.

The `editors/code/dada.tmLanguage.json` file does not fully cover the syntax yet, but at least we have some colors now. =)

I think it would be cool to have semantic highlighting as well. Then we would be able to highlight e.g., `Foo()` differently depending on if `Foo` is a function or a class.

Closes #5.


XFFXFF:
I tried it locally and it looks cool! 

image

nikomatsakis:
Neat! I've not had any brain space for Dada but I'll take a look at some point!

nikomatsakis:
@Aadamandersson Wow, neat! Sorry for not responding for all this time! I was snowed under with notifications and other things. Let me check this out!

nikomatsakis (approved):

PR #225, opened 2023-05-09
Fix offsets for LSP diagnostics

Aadamandersson:
This fixes the offsets for LSP diagnostics.
The `line` and `character` in `Position` is zero-based (https://docs.rs/lsp-types/0.94.0/src/lsp_types/lib.rs.html#231-240)


brson:
Thanks!

PR #226, opened 2023-05-10
Fix warnings in CI

Aadamandersson:
This fixes the warnings in our CI.
We are currently using  [action-rs/toolchain](https://github.com/actions-rs/toolchain) to install the Rust toolchain in Github Actions. But as it seems to be unmaintained (see actions-rs/toolchain#216) and needs to be replaced to fix some of the warnings (see actions-rs/toolchain#219) I have replaced it with [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain), which looks like a suitable replacement.

### DIFF
--- a/.github/bors.toml
+++ b/.github/bors.toml
@@ -1,9 +1,9 @@
 status = [
-  "Check (nightly)",
-  "Test (nightly)",
+  "Check",
+  "Test",
   "Rustfmt",
-  "Clippy (nightly)",
-  "Deploy (nightly)",
+  "Clippy",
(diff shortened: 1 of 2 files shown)


XFFXFF:
Do you mean such warnings?

image

XFFXFF (approved):
Looks good to me

Aadamandersson:
> Do you mean such warnings?
> 
> image

Yes, exactly

Aadamandersson:
Oh, I see. Thanks!

PR #227, opened 2023-05-11
Parse generic parameters

Aadamandersson:
This implements parsing of generic parameters for types `fn foo[T]` and permissions `fn foo[perm P]`.


nikomatsakis:
Nice! I had thought about having the syntax `fn foo[type T]`, the intention being that every declaration of a generic parameter is preceded by its kind -- but I hadn't made up my mind yet. 

One of the advantages would be that you can do interesting things with `impl` blocks:

```rust
class Foo[type T]() {
}

impl Foo[type T] {
    fn bar() { ... }
}

impl Foo[String] {
    ...
}
```

I'm not sure if this is the direction Dada should go, but I'd say it's worth giving it a try. 

Though I guess that would limit eventual Rust interop pretty dramatically.

nikomatsakis (approved):
Still, let's just merge it

PR #228, opened 2023-05-18
Fix error message for unavailable wasm-pack binary

Aadamandersson:
Running `cargo xtask deploy` on an operating system and architecture that does not have a wasm-pack binary available results in the following error:
Error: no wasm-pack binary available for `bool` and `bool`

This PR fixes the error message to actually print the os and arch, e.g.:
Error: no wasm-pack binary available for `macos` and `aarch64`


brson:
Thanks!

PR #238, opened 2025-03-17
Checked dada

nikomatsakis:
I think it's time to bring it down to mainline. It's a from-scratch reimplementation that is using the new Salsa and focused on the type system and WASM code generation. Still doesn't quite work and the tutorials etc are not done.


nikomatsakis:
OMG this is the most whiny CI ever

nikomatsakis:
MAKE IT STOP

PR #249, opened 2025-04-10
Emit wasm

brson:
Just a little tidbit of wasm, enough to compile and run an empty main function.

Fix the existing wasm codegen, including commenting out a non-working `local.set` for the stack pointer. I'm not sure what's supposed to happen with the stack pointer argument. Export the main function.

For now I've just added an `--emit-wasm` flag to `dada compile`, and not implemented `dada run`. I'm not sure if we should just have `dada run` call out to a wasmtime bin, use the wasmtime API or have the option for both.

~Add a `dada-run` crate that links to wasmtime as an initial host. I'd expect this crate to grow a lot.~

~Wire them up.~

~Next steps include adding some intrinsic printing functions and extending the test harness to run and check output. Maybe add a way to dump wasm and/or wat files. Suggestions welcome.~

### DIFF
--- a/components/dada-codegen/src/cx.rs
+++ b/components/dada-codegen/src/cx.rs
@@ -1,7 +1,7 @@
 use dada_ir_sym::{ir::functions::SymFunction, ir::types::SymGenericTerm};
 use dada_util::{FromImpls, Map};
 use salsa::Update;
-use wasm_encoder::{CodeSection, FunctionSection, TypeSection};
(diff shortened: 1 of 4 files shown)


brson:
I'm considering whether we can immediately skip having a runtime at all and just call the `wasmtime` bin to run dada programs.

brson:
I'll keep working on this.

